PR #5, opened 2015-03-11
use setuptools find_packages to find packages

graingert:
https://pythonhosted.org/setuptools/setuptools.html#using-find-packages


graingert:
https://pythonhosted.org/setuptools/setuptools.html#using-find-packages

adamchainz:
plz test next time it totally broke

PR #6, opened 2015-03-11
Fix find_packages typo

graingert:


graingert:
Woops, I thought you ran tox as part of your tests

adamchainz:
No, and no this still didn't work. I'm sticking with the tried and tested `get_packages` from django rest framework

Issue #9, opened 2015-03-19
Support handler API


adamchainz:
In my benchmarks, HANDLER can perform about twice as fast for reading tons of rows (10k random pk lookups). It could be useful for some situations to be directly supported on django models. Imagined API:

```python
with Author.objects.handler('pk') as handler:
    handler.read(5)
```

PR #15, opened 2015-03-21
Support HANDLER - #9

adamchainz:
Relatively simple API that supports the HANDLER statements of forms 1 and 2 only. Form 3, 'natural row' order, didn't seem worth supporting at this time since it's not a thing in InnoDB. (N.B. it may be *even faster* though for those non-transactional storage engines).


adamchainz:
Needs complete docs...

PR #16, opened 2015-03-21
Add Set fields

### LINKED ISSUE #10: Set fields
MySQL has `FIND_IN_SET`. This could be used to implement model fields that serialize to a set, as a poor man's parallel of the postgres array fields that `django.contrib.postgres` has.

adamchainz:
Resolves #10, as described there.


adamchainz:
`SimpleSetField` needs to warn about duplicates in `to_python` - it's not user friendly to allow "1,1" to be typed but of course a set only stores "{1}"

PR #55, opened 2015-04-01
Document and test deficiency of the index-based lookup on list fields	

adamchainz:
We're treating lists like sets, so if you try do `field__1="red"` it will not match `["red", "red"]`, since `FIND_IN_SET` returns that it's at the first position in the list.

Should document this or remove the positional lookup (since `contains` goes 90% of the way and does not carry any risk of misunderstanding).


adamchainz:
Perhaps leaving the positional lookup in with great documentation about its supreme deficiencies would be enough...

PR #57, opened 2015-04-03
pt_visual_explain doesn't execute the query to capture it

### LINKED ISSUE #56: pt_visual_explain need not execute the query
There must be some way of monkey-patching somewhere in MySQLdb/mysqlclient or django to grab the full query from the queryset without executing it. Find it!

adamchainz:
Resolves #56 . Gets the query's SQL and params, prepends 'explain', and executes it, to get the fully param-substituted form, before passing THAT to mysql + pt-visual-explain

### DIFF
--- a/django_mysql/models/query.py
+++ b/django_mysql/models/query.py
@@ -1,5 +1,5 @@
 # -*- coding:utf-8 -*-
-from __future__ import print_function
+from __future__ import print_function, unicode_literals
 
 import sys
 from copy import copy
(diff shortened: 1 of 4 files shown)


adamchainz:
To do:

1. Extract the query formatter to a utils function
2. Make it work with oracle's MySQL connector/python ? (The only other MySQL DBAPI library suggested by django - https://docs.djangoproject.com/en/1.7/ref/databases/#mysql-db-api-drivers ). It's harder to add tests since Oracle refuse to put it on PYPI.

adamchainz:
My bad, turns out it is "on PyPI", but externally hosted, as this works: `pip install --allow-external mysql-connector-python mysql-connector-python`. I think I misunderstood when I heard "it's not on PyPI".

Issue #52, opened 2015-03-30
Common_schema integration


adamchainz:
`common_schema` provides some interesting integration possibilities. More research required - it may just be too DBA/SQL centric to make sense for app code, but worth consideration. http://common-schema.googlecode.com/svn/trunk/common_schema/doc/html/introduction.html

adamchainz:
Particularly useful:

* date functions - http://common-schema.googlecode.com/svn/trunk/common_schema/doc/html/temporal_routines.html
* Maybe a few text functions like trim_wspace - http://common-schema.googlecode.com/svn/trunk/common_schema/doc/html/text_routines.html
* crc64 etc: http://common-schema.googlecode.com/svn/trunk/common_schema/doc/html/general_routines.html
* throttle ? http://common-schema.googlecode.com/svn/trunk/common_schema/doc/html/process_routines.html

adamchainz:
After investigation during creating #65 , found not many of the functions would be that useful - they're a bit too script-oriented really. This could be re-opened if someone asks for it though.

PR #65, opened 2015-04-15
Add CRC64 function

adamchainz:
From #52 - copied the definition out from common schema. Can be more useful than CRC32 for group-by hashing etc. so keeping it!


adamchainz:
Can't actually think of a use case and until I see one for CRC32 then I won't add CRC64

PR #82, opened 2015-04-26
Travis tests MariaDB 5.5 again

adamchainz:
The apt repo seems to have broken. Keeping this open as a pull request and hitting rebuild on travis til it works again - otherwise more drastic changes may be required.


adamchainz:
Looks like the repo is finally working again.

PR #83, opened 2015-04-28
Add DynamicField for MariaDB Dynamic Columns

adamchainz:
Todo:

- [x] create/update/delete
- [x] Dynamic column functions as `Function`s
- [x] Copied and adapted [custom lookups](https://docs.djangoproject.com/en/1.8/ref/contrib/postgres/fields/#querying-hstorefield) from `django.contrib.postgres` `HStoreField` and `JSONField`
- [x] Field Docs
- [x] Function Docs

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,8 @@ History
 -----------------------
 
 * Pending release
+* Added ``DynamicField`` for using MariaDB's Named Dynamic Columns, and related
+  database functions ``ColumnAdd``, ``ColumnDelete``, and ``ColumnGet``.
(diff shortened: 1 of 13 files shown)


adamchainz:
Using https://github.com/django/django/pull/5099 as a base, can test with `tox -e py27-djangoinsertexpressions,py34-djangoinsertexpressions`. Looks like this will be 1.9+ only - unless there is a (relatively easy) way to format the dynamic columns blob locally.

adamchainz:
N.b. since creating https://github.com/adamchainz/mariadb-dyncol have been pushing ahead for 1.7+ compatible solution with formatting in python

PR #91, opened 2015-05-05
Add SizedBinaryField and SizedTextField

### LINKED ISSUE #76: Small / medium / large fields
Django's fields all prescribe one size for things, but maybe sometimes you want a `TINYTEXT` or a `TINYBLOB` to save some bytes. Also maybe you have a legacy application that uses these that you want to accurately represent in djangoland.

These are easy subclasses of the built-in fields with just the `db_type` changed, but there should be some place to import them from (`django_mysql.fields` !) rather than be forced to do the subclassing yourself in your app. Making a field class isn't as trivial as it could be.

adamchainz:
Part of #76 , allows TINYBLOB, BLOB, MEDIUMBLOB, etc. to be used in Django.

### DIFF
--- a/django_mysql/models/fields/__init__.py
+++ b/django_mysql/models/fields/__init__.py
@@ -1,2 +1,3 @@
 from django_mysql.models.fields.lists import *  # NOQA
 from django_mysql.models.fields.sets import *  # NOQA
+from django_mysql.models.fields.sizes import *  # NOQA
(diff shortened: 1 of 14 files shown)


adamchainz:
Idea - if these were implemented as a single field subclassing e.g. `BinaryField`, with one extra argument for the size, could `deconstruct` be overridden and the field made to work with migrations of increasing/decreasing size, as a bonus?

adamchainz:
Answer to previous question: yes. Patch re-rewritten.

PR #123, opened 2015-06-01
Simple fields customizable

### LINKED ISSUE #122: Make SimpleListField and SimpleSetField easier to customize
Some use cases of comma separated fields are better off displaying the data with newlines, or as json, etc. It's not so easy right now to subclass `SimpleListField` or `SimpleSetField` right now to customize just the [de]serialization logic whilst keeping the validation logic - make it so, by moving the string splitting/combining into separate methods.

adamchainz:
Resolves #122 .

Todo:

- [x] Make them customizable
- [x] Test fully
- [ ] Docs

### DIFF
--- a/django_mysql/forms.py
+++ b/django_mysql/forms.py
@@ -18,7 +18,8 @@ __all__ = ('SimpleListField', 'SimpleSetField')
 
 class SimpleListField(forms.CharField):
 
-    # These bits can be overridden simply to change the way the field works
+    # These bits can be overridden to change the way the field serializes and
+    # deserializes for the user, e.g. line-delimited, json, etc.
 
     default_error_messages = {
         'item_n_invalid': _('Item %(nth)s in the list did not validate: '),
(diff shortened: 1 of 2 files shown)


adamchainz:
Could also just support the `delimiter` argument like https://docs.djangoproject.com/en/1.8/ref/contrib/postgres/forms/ - although it's a little less flexible and requires the default error messages to change to be more generic.

adamchainz:
I'm not entirely sold on this. `django.contrib.postgres`'s SimpleArrayField is more customizable but this is mostly because the delimiter is not part of the database backend. In most cases I think it's probably better for the user to copy/paste the class and rewrite the validation logic because it's a bit tangled - realistically this pull request will only allow the customization of the delimiter and maybe the addition of simple things like `.strip()` on the input, which I don't think is really worth it.

adamchainz:
Abandoning after re-reasoning the above.

PR #127, opened 2015-06-08
Add TimestampField

adamchainz:
Todo
- [ ] Basic tests
- [ ] `DEFAULT CURRENT_TIMESTAMP`, `ON UPDATE CURRENT_TIMESTAMP`
- [ ] Migration tests
- [ ] Docs
- [ ] Release note

### DIFF
--- a/django_mysql/models/fields/__init__.py
+++ b/django_mysql/models/fields/__init__.py
@@ -2,3 +2,4 @@ from django_mysql.models.fields.bit import *  # NOQA
 from django_mysql.models.fields.lists import *  # NOQA
 from django_mysql.models.fields.sets import *  # NOQA
 from django_mysql.models.fields.sizes import *  # NOQA
+from django_mysql.models.fields.timestamp import *  # NOQA
--- a/django_mysql/models/fields/timestamp.py
+++ b/django_mysql/models/fields/timestamp.py
@@ -1,0 +1,39 @@
+# -*- coding:utf-8 -*-
+from datetime import datetime
+
+from django.db.models import DateTimeField
+
+from django_mysql.compat import field_class
+
+__all__ = ('TimestampField',)
+
+
+class TimestampField(field_class(DateTimeField)):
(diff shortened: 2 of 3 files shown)


adamchainz:
I don't think this is worth it

StevenMapes:
May I ask why you didn't think this was worth it? If you are using a database that can be written to from outside of django then being ale to add this in may be a requirement to allow those other sources records to work as the database engineer intended.

I've actually had this for clients where I have ported existing databases into django, managed to convince them to allow me to manage them from within django but where they still writin records from outside of the python project. In this case I have created a new new subclass of DatetTimeField directly within the project and overloaded the db_type method to set these up checking the connection type was MySQL/MariaDB first of all

```
class DateTimeWithDefaultAndUpdateCurrentTimeStampField(models.DateTimeField):
    """DateTimeField Extension to add in CURRENT_TIMESTAMP default behaviour"""
    def db_type(self, connection):
        if connection.vendor == "mysql":
            typ = ['DATETIME']
            typ += ['default CURRENT_TIMESTAMP() ON UPDATE CURRENT_TIMESTAMP()']
            return ' '.join(typ)
        else:
            return super(DateTimeWithDefaultAndUpdateCurrentTimeStampField, self).db_type(connection)


class DateTimeDefaultCurrentTimeStampField(models.DateTimeField):
    """DateTimeField Extension to add in CURRENT_TIMESTAMP default behaviour"""
    def db_type(self, connection):
        if connection.vendor == "mysql":
            typ = ['DATETIME']
            typ += ['default CURRENT_TIMESTAMP()']
            return ' '.join(typ)
        else:
            return super(DateTimeDefaultCurrentTimeStampField, self).db_type(connection)
```

adamchainz:
Maybe it would be worth it

adamchainz:
I don't have time to work on this. If there was a PR for a complete implementaiton, sure I'd review and merge it. This PR can serve as a starting point.

PR #144, opened 2015-08-06
No QuerySet.extra on Django 1.8+

adamchainz:
As discussed on [django developers mailing list](https://groups.google.com/forum/#!topic/django-developers/FojuU0syO8Y), `.extra()` is deprecated-ish and the only use of it in **django-mysql** shoudl be replaced with `RawSQL`. However it seems this is not possible, since we need a `RawQ` to just add `WHERE 1`.

### DIFF
--- a/django_mysql/compat.py
+++ b/django_mysql/compat.py
@@ -8,7 +8,8 @@ import functools
 import django
 from django.utils import six
 
-__all__ = ('BaseExpression', 'Func', 'Value', 'field_class')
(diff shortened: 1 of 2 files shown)


adamchainz:
After several attempts, it appears this is not easy still. You can't pass arbitrary objects to `filter` - they get inspected for the field names that are referenced. My last idea is to pass some kind of `filter(pk=FunctionThatDoesntUsePk('thecomment')` but this would still influence the query as to what gets selected - the pk...

adamchainz:
I don't think this will be possible for some time

PR #150, opened 2015-09-06
Add sql_calc_found_rows()

### LINKED ISSUE #115: SQL_CALC_FOUND_ROWS
With the work for query hints in place from #114 , it wouldn't be hard to add in [SQL_COUNT_FOUND_ROWS](https://mariadb.com/kb/en/mariadb/select/#sql_calc_found_rows). This could then be used in conjunction with the database function [FOUND_ROWS()](https://mariadb.com/kb/en/mariadb/found_rows/) to override Django's Paginator class to count the rows once - if you check [its source](https://github.com/django/django/blob/master/django/core/paginator.py) it doesn't seem too hard, as only `_get_count` needs overriding to return the count found when taking the slice of `object_list` in `_get_page`.

adamchainz:
Resolves #115 

To do:

- [x] `sql_calc_found_rows`
- [x] docs

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,6 +9,11 @@ History
 * Added index query hints to ``QuerySet`` methods, via query-rewriting layer
 * Add ``ordering`` parameter to ``GroupConcat`` to specify the ``ORDER BY``
   clause
+* Pending release
+* Added index query hints to ``QuerySet`` methods, via qurey-rewriting layer
+* Added ``sql_calc_found_rows()`` query hint that calculates the total rows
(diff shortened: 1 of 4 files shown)


adamchainz:
The paginator turned out to be very hard to do, since many of the `Paginator` features depend on fetching the `count` before ever slicing the `QuerySet`. Potentially a re-implementation of paginator with only the supported features would work, but subclassing was becoming a nightmare due to state of the object. I've left the code in the [found_rows_paginator](https://github.com/adamchainz/django-mysql/tree/found_rows_paginator) branch.

PR #154, opened 2015-09-22
information_schema as models

### LINKED ISSUE #153: information_schema tables as models
The various `information_schema` tables could be expressed as Django models. They would not trigger migrations with with `managed = False` and could be importable as e.g. `django_mysql.models.information_schema`, for `information_schema.Schemata` etc. The only two hacks I can forsee would be:
1. The `db_table` param would really need to contain the database and table i.e. `information_schema.tablename`
2. The `objects` managers on many of them should be automatically filtered to ones' own database only by default.

adamchainz:
Resolves #153 .

Todo:

- [ ] all the tables
- [ ] all the docs


adamchainz:
I can't think of many usecases atm, closing for now

PR #175, opened 2015-11-04
Test on Python 3.5

adamchainz:


adamchainz:
Still missing from Travis :(

adamchainz:
Was fixed in #181

PR #180, opened 2015-11-07
Support chunk_min in SmartChunkedIterator

lwiecek:
Fixes #180

### DIFF
--- a/django_mysql/models/query.py
+++ b/django_mysql/models/query.py
@@ -231,8 +231,8 @@ class ApproximateInt(int):
 
 class SmartChunkedIterator(object):
     def __init__(self, queryset, atomically=True, status_thresholds=None,
-                 pk_range=None, chunk_time=0.5, chunk_size=2, chunk_max=10000,
-                 report_progress=False, total=None):
+                 pk_range=None, chunk_time=0.5, chunk_size=2, chunk_min=1,
+                 chunk_max=10000, report_progress=False, total=None):
         self.queryset = self.sanitize_queryset(queryset)
 
         if atomically:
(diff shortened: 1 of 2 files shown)


django_mysql/models/query.py
```diff
@@ -246,8 +246,11 @@ def __init__(self, queryset, atomically=True, status_thresholds=None,
         self.pk_range = pk_range
 
         self.rate = WeightedAverageRate(chunk_time)
+        assert 0 < chunk_min <= chunk_max, \
```
adamchainz: Don't like line continuations, use the syntax:

```
assert (
    0 < chunk_min <= chunk_max,
    "Minimum...
)
```

django_mysql/models/query.py
```diff
@@ -246,8 +246,11 @@ def __init__(self, queryset, atomically=True, status_thresholds=None,
...
+            "Minimum chunk size should not be greater than maximum chunk size."
+        self.chunk_min = chunk_min
         self.chunk_max = chunk_max
-        self.chunk_size = min(chunk_size, chunk_max)
+        self.chunk_size = self.within_min_and_max(chunk_size)
```
adamchainz: the method name is a bit explicit, why not call it `self.constrain_size`?

tests/testapp/test_models.py
```diff
@@ -530,6 +530,22 @@ def test_objects_max_size_bounds_chunk_size(self):
...
+        for chunk in smart:
+            ids = [author.id for author in chunk]
+            assert len(ids) >= 5
+            seen.extend(ids)
+        all_ids = list(Author.objects.order_by('id')
+                                     .values_list('id', flat=True))
+        assert seen == all_ids
+
+    def test_min_size_greater_than_max(self):
+        with self.assertRaises(AssertionError):
```
adamchainz: use

```
with pytest.raises(AssertionError):
```

django_mysql/models/query.py
lwiecek: assert doesn't like parentheses:
```
>>> assert(False, "abc")
<stdin>:1: SyntaxWarning: assertion is always true, perhaps remove parentheses?
```

adamchainz: It does if you leave a space

adamchainz:
Please squash and retitle "Support chunk_min in SmartChunkedIterator".

adamchainz:
Merged in 01437dd1f09ceb5b6f19c68d91a136fc591e682f

PR #182, opened 2015-11-07
Tests run against Django master too

adamchainz:
Broken until https://code.djangoproject.com/ticket/20846#comment:25 has a response.


adamchainz:
That was fixed in django/django@780bddf75b93784470a2e352ed44ee35a751d667 , an update shows just some argument parsing fixes to implement atm.

PR #185, opened 2015-11-10
Add EnumField

### LINKED ISSUE #140: Add EnumField
MySQL's ENUM type has no Django field apart from [on Stack Overflow](http://stackoverflow.com/questions/21454/specifying-a-mysql-enum-in-a-django-model). This is a shame for people using legacy DBs.

Similar to #136 and #91, we should add an EnumField that allows using these columns. A bonus would be to get migrations going smoothly such that adding/removing a single enum value would go well - the `ALTER TABLE` should be instant.

MichaelAquilina:
Fixes #140 .
As requested, I'm opening this pull request early so that we can discuss any design details on github. I'll probably get a chance to continue working on this again on Wednesday or Thursday.

### DIFF
--- a/django_mysql/models/__init__.py
+++ b/django_mysql/models/__init__.py
@@ -8,7 +8,7 @@ from django_mysql.models.query import (  # noqa
     pt_visual_explain, QuerySet, QuerySetMixin
 )
 from django_mysql.models.fields import (  # noqa
-    Bit1BooleanField, DynamicField, ListCharField, ListTextField,
+    Bit1BooleanField, DynamicField, EnumField, ListCharField, ListTextField,
     NullBit1BooleanField, SetCharField, SetTextField, SizedBinaryField,
     SizedTextField,
 )
--- a/django_mysql/models/fields/enum.py
+++ b/django_mysql/models/fields/enum.py
@@ -1,0 +1,59 @@
+# -*- encoding:utf-8 -*-
+from __future__ import absolute_import, division, unicode_literals
+
+import six
+from django.db.models import CharField
+from django.utils.translation import ugettext_lazy as _
+
+from django_mysql.compat import field_class
+
+from _mysql import escape_string  # isort:skip
+
+
+class EnumField(field_class(CharField)):
+    description = _("MySQL ENUM Type")
+
+    def __init__(self, *args, **kwargs):
+        if 'choices' not in kwargs or len(kwargs['choices']) == 0:
+            raise ValueError(
+                '"choices" argument must be be a non-empty list'
+            )
+
+        choices = []
+        for choice in kwargs['choices']:
+            if isinstance(choice, tuple):
+                choices.append(choice)
+            elif isinstance(choice, six.string_types):
+                choices.append((choice, choice))
+            else:
+                raise TypeError(
+                    'Invalid choice "{choice}". '
+                    'Expected string or tuple as elements in choices'.format(
+                        choice=choice,
+                    )
+                )
+
+        kwargs['choices'] = choices
+        super(EnumField, self).__init__(*args, **kwargs)
+
+    def _check_max_length_attribute(self, **kwargs):
+        # Disable max length check on Django 1.7
+        return []
+
+    def deconstruct(self):
+        name, path, args, kwargs = super(EnumField, self).deconstruct()
+        bad_paths = (
+            'django_mysql.models.fields.enum.' + self.__class__.__name__,
+            'django_mysql.models.fields.' + self.__class__.__name__
+        )
+        if path in bad_paths:
+            path = 'django_mysql.models.' + self.__class__.__name__
+
+        kwargs['choices'] = self.choices
+        return name, path, args, kwargs
+
+    def db_type(self, connection):
+        values = [escape_string(c) for c, _ in self.flatchoices]
+        return 'enum(%s)' % ','.join(
+            "'%s'" % v.decode('utf8') for v in values
+        )
--- a/docs/exposition.rst
+++ b/docs/exposition.rst
@@ -161,6 +161,18 @@ string::
 
 :ref:`Read more <set-fields>`
 
+Enum Fields
+-----------
+
+You can take advantage of MySQL's ``ENUM`` type through the ``EnumField``
+provided by django-mysql. Using ``ENUM`` provides the benefit of storing
+recurring strings in a very space efficient manner.
+
+    class BookCover(Model):
+        color = EnumField(choices=['red', 'green', 'blue'])
+
+:ref:`Read more <enum-fields>`
+
 
 Resizable Text/Binary Fields
 ----------------------------
--- a/docs/model_fields.rst
+++ b/docs/model_fields.rst
@@ -695,6 +695,43 @@ Or with attribute assignment to a model::
         any conflicts.
 
 
+.. _enum-fields:
+
+-----------
+Enum Fields
+-----------
+
+Using a ``CharField`` with a limited set of strings leads to inefficient data
+storage since the string value must be stored over and over. MySQL's ``ENUM``
+type allows a more compact representation of such columns by storing the list
+of strings just once and using integers in the data rows, which ``EnumField``
+allows you to use conveniently::
+
+    class BookCover(Model):
+        color = EnumField(choices=['red', 'green', 'blue'])
+
+Like Django's standard ``choices`` argument, the values may be specified as a
+tuple specifying the actual value and a human readable name. The actual value
+is what will be stored in the database, while the human readable value will be shown
+in Django Admin forms where applicable::
+
+    class Book(Model):
+        color = EnumField(choices=[
+          ('red', 'Bright Red'),
+          ('green', 'Vibrant Green'),
+          'blue',  # human readable name will be set to "blue"
+        ])
+
+
+.. note::
+   It is possible to introduce new enumeration choices in a migration, as well as
+   edit the *human readable* names of current choices.
+
+   However, editing or removing existing choice values will result in corresponding
+   rows to be replaced with the empty string. Altering a table in such a manner will
+   only be allowed if MySQL Strict Mode is not on.
+
+
 .. _resizable-blob-text-fields:
 
 ----------------------------
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -11,3 +11,4 @@ Contributors
 ------------
 
 * Łukasz Więcek <https://github.com/lwiecek>
+* Michael Aquilina <https://github.com/MichaelAquilina>
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,6 +9,7 @@ Pending
 * New release notes here
 * New function ``add_QuerySetMixin`` allows addding the ``QuerySetMixin`` to
   arbitrary ``QuerySet``\s, for when you can't edit a model class.
+* Added field class ``EnumField`` that uses MySQL's ``ENUM`` data type.
 
 1.0.1 (2015-11-18)
 ------------------
--- a/django_mysql/models/fields/__init__.py
+++ b/django_mysql/models/fields/__init__.py
@@ -2,6 +2,7 @@ from django_mysql.models.fields.bit import (  # noqa
     Bit1BooleanField, NullBit1BooleanField
 )
 from django_mysql.models.fields.dynamic import DynamicField  # noqa
+from django_mysql.models.fields.enum import EnumField  # noqa
 from django_mysql.models.fields.lists import (  # noqa
     ListCharField, ListTextField
 )
--- a/tests/testapp/enum_default_migrations/0001_initial.py
+++ b/tests/testapp/enum_default_migrations/0001_initial.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from django_mysql.models import EnumField
+
+
+class Migration(migrations.Migration):
+
+    dependencies = []
+
+    operations = [
+        migrations.CreateModel(
+            name='EnumDefaultModel',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False,
+                 auto_created=True, primary_key=True)),
+                ('field', EnumField(choices=[
+                    ('lion', 'Lion'), ('tiger', 'Tiger'), 'bear'
+                ])),
+            ],
+            options={
+            },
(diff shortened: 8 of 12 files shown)


django_mysql/models/__init__.py
```diff
@@ -8,7 +8,7 @@
     QuerySet, QuerySetMixin
 )
 from django_mysql.models.fields import (  # noqa
-    Bit1BooleanField, DynamicField, ListCharField, ListTextField,
+    Bit1BooleanField, DynamicField, EnumField, ListCharField, ListTextField,
     NullBit1BooleanField, SetCharField, SetTextField, SizedBinaryField,
-    SizedTextField,
```
MichaelAquilina: i'll re-add that trailing comma

adamchainz:
:rainbow: :rabbit: :rabbit2: :dog: :poodle: :dog2: :shipit:

django_mysql/models/__init__.py
adamchainz: :rage4:

tests/testapp/test_dynamicfield.py
```diff
@@ -316,6 +316,7 @@ def test_key_transform_nesty__level2__startswith(self):
             [self.objs[4]]
         )
 
+# You should be using pytest skipifs!
```
adamchainz: no i shouldn't

MichaelAquilina: That wasn't meant to be included in this diff :p I wrote it before I discussed it with you

MichaelAquilina:
Interesting that Django 1.7 is failing. I'm pretty certain I know what the issue is though

adamchainz:
Todo:

1. Add a field check that ensures that on Python < 3.4, `enum34` is installed and errors if this is not the case
2. Add a field check that ensures that `enum` is an `Enum` from `enum34` or Python core and errors if this is not the case
3. Add a field check that hits the db (see DynamicField - you have to restrict it to 1.8 + and use 'nodb_connection') and checks that the comma-separated variable `sql_mode` includes `STRICT_TRANS_TABLES` or `STRICT_ALL_TABLES` and emits a *warning* if this is not the case
4. Add a field check that `enum` is not greater than the max MySQL size (see manual, it's something like 65535) and errors if it is
5. Add a field that there are no duplicates in `enum` (is this necessary?)
6. Test migrations, especially with adding and removing. If possible, make it an error to remove enum values after they have been added... ?

MichaelAquilina:
Do we really want to make a check that enum34 is installed if python2 is being used? We can set it as a conditional requirement for python 2 installations.

If not, then if the user does not have eum34 installed, then they will not be able pass a valid instance to EnumField (which we can document).

adamchainz:
It can't be a conditional requirement because users should be able to use
Django-MySQL without using enumfield. Also the check that the enum quacks
like an enum should be done anyway.

On Friday, 13 November 2015, Michael Aquilina <notifications@github.com>
wrote:

> Do we really want to make a check that enum34 is installed if python2 is
> being used? We can set it as a conditional requirement for python 2
> installations.
>
> If not, then if the user does not have eum34 installed, then they will not
> be able pass a valid instance to EnumField (which we can document).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/adamchainz/django-mysql/pull/185#issuecomment-156467280>
> .
>


-- 
Adam

MichaelAquilina:
Discussion Summary:
-----------------------------

Change back to supporting strings instead of Enum (extend CharField). This is because of numerous issues with Enum types such as not supporting strings like "example.com" and "hello@example" in a way that is easy to use. 

Using Enum also complicates the code significantly and requires us having to write lookups for a lot of stuff already built into CharField.

We will use choices argument to specify the enum strings:
```
EnumField(choices=[
   ('foo', 'Foo'),
   ('bar', 'Bar'),
])
```
We will also allow specifying values as follows:
```
EnumField(choices=['Foo', 'Bar'])
```
this is easily done but expanding it to a tuple ourselves in `__init__`

We will use `flatchoices` internally to construct the enum definition in the database (see `Field` in the django code base)

see: https://docs.djangoproject.com/en/1.8/ref/models/fields/

adamchainz:
Other things:

* Copy and test the path normalization logic found in `deconstruct` on the other `django_mysql` fields
* We need to test migrations for both adding and removing values from `choices`
* Make `choices` required and non-empty (raises error on MySQL) with something like:

```
def __init__(**kwargs):
    choices = kwargs.pop('choices', None)
    if not choices:
        raise ValueError("'choices' must be filled in, and cannot be the empty list")
```

MichaelAquilina:
Rebased + other some tests (including migration tests)

django_mysql/models/fields/enum.py
```diff
@@ -0,0 +1,53 @@
...
+            else:
+                raise TypeError(
+                    'Expected string or tuple as elements in choices'
+                )
+
+        kwargs['choices'] = choices
+        super(EnumField, self).__init__(*args, **kwargs)
+
+    def _check_max_length_attribute(self, **kwargs):
+        # Required for django1.7
```
adamchainz: Change to say "Disable max length check on Django 1.7"

adamchainz:
Todo:

- [ ] 100% coverage - test the assertions (both branches) about type of `choices`
- [ ] Release note
- [ ] Docs
- [ ] Squash into single commit

django_mysql/models/fields/enum.py
```diff
@@ -0,0 +1,53 @@
...
+
+from django_mysql.compat import field_class
+
+
+class EnumField(field_class(CharField)):
+    description = _("MySQL ENUM Type")
+
+    def __init__(self, *args, **kwargs):
+        if 'choices' not in kwargs or len(kwargs['choices']) == 0:
+            raise TypeError(
```
adamchainz: I think `ValueError` is better

django_mysql/models/fields/enum.py
```diff
@@ -0,0 +1,53 @@
...
+
+        choices = []
+        for choice in kwargs['choices']:
+            if isinstance(choice, tuple):
+                choices.append(choice)
+            elif isinstance(choice, six.string_types):
+                choices.append((choice, choice))
+            else:
+                raise TypeError(
+                    'Expected string or tuple as elements in choices'
```
adamchainz: try print the bad `choice` in the message

MichaelAquilina:
Got the following left to do:

* Docs
* Squash

MichaelAquilina:
Everything squashed. Now just docs left

MichaelAquilina:
Ready for another round of review!

docs/exposition.rst
```diff
@@ -161,6 +161,29 @@ string::
...
+is what will be stored in the database, while the human readable value will be shown
+in Django Admin forms where applicable::
+
+    class Book(Model):
+        color = EnumField(choices=[
+          ('red', 'Bright Red'),
+          ('green', 'Vibrant Green'),
+          'blue',  # human readable name will be set to "blue"
+        ])
+
```
adamchainz: Add a note about migrating: first explain that it's possible, second that you shouldn't remove one ever, since everything gets renumbered.

adamchainz: Also link to both MySQL 5.5 docs and MariaDB docs for ENUM

docs/exposition.rst
```diff
@@ -161,6 +161,29 @@ string::
 
 :ref:`Read more <set-fields>`
 
+Enum Fields
+-----------
+
+You can take advantage of MySQL's ``ENUM`` type through the ``EnumField``
+provided by django-mysql. Specifying the range of values available
```
adamchainz: explain why one woudl use ENUM over CharField with choices - massive space saving due to underlying int, plus restrained set of strings so you know there is no bad data

django_mysql/models/fields/enum.py
```diff
@@ -0,0 +1,56 @@
...
+        if path in bad_paths:
+            path = 'django_mysql.models.' + self.__class__.__name__
+
+        kwargs['choices'] = self.choices
+        return name, path, args, kwargs
+
+    def db_type(self, connection):
+        return 'enum(%s)' % ','.join(
+            "'%s'" % v for v in [c for c, _ in self.flatchoices]
+        )
```
adamchainz: I don't think this does escaping correctly... what if you change one of your migratiosn to a string that includes the backtick character `

docs/exposition.rst
MichaelAquilina: How would you like t-the link to be specified? Cant seem to find any other external references you used to mysql/mariadb in the exposition documentation

django_mysql/models/fields/enum.py
MichaelAquilina: Fair point. I'll fix that and add a test

docs/exposition.rst
adamchainz: I missed that this was in the exposition... the exposition is only for short summaries, you need to write main documentation in `model_fields.rst`. I think you should move what you have over and then try write a shorter summary here, like the pair of examples used for List Fields.

MichaelAquilina:
So django 1.7 does not seem to provide a reference to the internal connection object that provides the `escape_string` according to the travis failures. Are you aware of any elegant ways round it?

docs/exposition.rst
```diff
@@ -161,6 +161,18 @@ string::
...
 
+Enum Fields
+-----------
+
+You can take advantage of MySQL's ``ENUM`` type through the ``EnumField``
+provided by django-mysql. Using ``ENUM`` provides the benefit of storing
+recurring strings in a very space efficient manner.
+
+    class BookCover(Model):
+        color = EnumField(choices=['red', 'green', 'blue'])s
```
adamchainz: trailing 's'

docs/model_fields.rst
```diff
@@ -695,6 +695,45 @@ Or with attribute assignment to a model::
...
+
+    class Book(Model):
+        color = EnumField(choices=[
+          ('red', 'Bright Red'),
+          ('green', 'Vibrant Green'),
+          'blue',  # human readable name will be set to "blue"
+        ])
+
+
+.. NOTE::
```
adamchainz: lowercase `note` ?

docs/model_fields.rst
```diff
@@ -695,6 +695,45 @@ Or with attribute assignment to a model::
...
+-----------
+
+You can take advantage of MySQL's ``ENUM`` type through the ``EnumField``
+provided by django-mysql. Using ``ENUM`` provides the benefit of storing
+recurring strings in a very space efficient manner. MySQL will store the
+enumeration value in each row as an integer value and return results as
+the corresponding strings in queries.
+
+Specifying the range of values available in an EnumField must be specified
+in the `choices` argument as a tuple or list::
```
adamchainz: escaping is with two backticks in RST. Have you managed to compile the docs locally? `cd docs` , `make html` or `make livehtml`

docs/exposition.rst
```diff
@@ -161,6 +161,18 @@ string::
 
 :ref:`Read more <set-fields>`
 
+Enum Fields
+-----------
+
+You can take advantage of MySQL's ``ENUM`` type through the ``EnumField``
+provided by django-mysql. Using ``ENUM`` provides the benefit of storing
```
adamchainz:
```
Using a ``CharField`` with a limited set of strings leads to inefficient data storage since the string value must be stored over and over. MySQL's ``ENUM`` type allows a more compact representation of such columns by storing the list of strings just once and using integers in the data rows, which ``EnumField`` allows you to use conveniently::
```

docs/model_fields.rst
```diff
@@ -695,6 +695,45 @@ Or with attribute assignment to a model::
...
+          'blue',  # human readable name will be set to "blue"
+        ])
+
+
+.. NOTE::
+   It is possible to introduce new enumeration choices in a migration, as well as
+   edit the *human readable* names of current choices.
+
+   However, editing or removing existing choice values will result in corresponding
+   rows to be represented as empty strings.
```
adamchainz: To be replaced with the empty string? Also make a note it's only possible to ``ALTER TABLE`` to truncate the data if MySQL Strict Mode is not on.

docs/model_fields.rst
```diff
@@ -695,6 +695,45 @@ Or with attribute assignment to a model::
         any conflicts.
 
 
+.. _enum-fields:
+
+-----------
+Enum Fields
+-----------
+
+You can take advantage of MySQL's ``ENUM`` type through the ``EnumField``
```
adamchainz: Copy the blurb above, no need to mention django-mysql.

docs/model_fields.rst
```diff
@@ -695,6 +695,45 @@ Or with attribute assignment to a model::
...
+provided by django-mysql. Using ``ENUM`` provides the benefit of storing
+recurring strings in a very space efficient manner. MySQL will store the
+enumeration value in each row as an integer value and return results as
+the corresponding strings in queries.
+
```
adamchainz: you should

1. add docs links
2. add class definition

```
 .. class:: EnumField(choices, **kwargs)
```

See below in the file for resizable binary/string fields examples.

docs/model_fields.rst
```diff
@@ -695,6 +695,45 @@ Or with attribute assignment to a model::
...
+Specifying the range of values available in an EnumField must be specified
+in the `choices` argument as a tuple or list::
+
+    class BookCover(Model):
+        color = EnumField(choices=['red', 'green', 'blue'])
+
+Like Django's standard `choices` argument, the values may be specified as a
```
adamchainz: It's possible to link into Django's dosc with intersphinx, if you manage to find the lookup for 'choices' I'd be impressed.

adamchainz:
Re: failures - it's possible to grab `escape_string` with direct importing:
```
from _mysql import escape_string
```

This will do for now I think.

MichaelAquilina:
The linter cant seem to decide what import it prefers for `from _mysql import escape_string` :(

adamchainz:
Put it with the second party imports and prevent sorting with `# isort:skip` ( https://github.com/timothycrosley/isort#skip-processing-of-imports-outside-of-configuration ).

docs/model_fields.rst
MichaelAquilina: I've always tested locally but never noticed choices being incorrect actually

MichaelAquilina:
I think everything should be ready for another round of review

adamchainz:
Looking good. I'm going to merge it with minor changes.

adamchainz:
Merged in 91ba48b1220d587b2f498502c07304fc98b49667, thanks!

PR #191, opened 2015-11-20
Docs - Ubuntu setup uses whoami to automate user creation

MichaelAquilina:


adamchainz:
Update the commit's message to be "Docs - Ubuntu setup uses whoami to automate user creation"

PR #226, opened 2016-02-05
Add fix_datetime_columns management command

### LINKED ISSUE #225: Command to output sql for converting datetime to datetime(6)
Loop over all django models with `DateTimeField`s and see if they are `datetime` in the database. If so, output the `ALTER TABLE` statement required to convert them to `datetime(6)`.

adamchainz:
Fixes #225. This command scans the database and outputs the SQL necessary to fix any `datetime` columns into `datetime(6)` columns, as is necessary when upgrading to MySQL 5.6, or MariaDB 5.3+.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,9 @@ Pending
 -------
 
 * New release notes here
+* Added ``manage.py`` command ``fix_datetime_columns`` that outputs the SQL
+  necessary to fix any ``datetime`` columns into ``datetime(6)``, as required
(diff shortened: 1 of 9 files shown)


adamchainz:
Okay, coverage has decreased because MySQL 5.6 is no longer being installed on Travis 5.6 environment, only 5.5! Looks like the PPA has gone bad. Should fix the build by merging in the work done to take MySQL from Oracle's repos in #155 !

However the tests pass locally for me on 5.7, so I'm assuming this is all good.

PR #242, opened 2016-02-23
Add lock and unlock methods to Lock/TableLock classes

lwiecek:
Summary: Refactored `Lock` a bit so `lock` / `unlock` can be used without `with`-statement

Checklist:

- [x] Docs
- [x] HISTORY.rst
- [x] Tests

Test Plan: `./runtests.py`

### DIFF
--- a/django_mysql/locks.py
+++ b/django_mysql/locks.py
@@ -41,7 +41,7 @@ class Lock(object):
         return self.lock()
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.unlock(exc_type, exc_value, traceback)
+        self.unlock()
 
     def lock(self):
         with self.get_cursor() as cursor:
@@ -58,7 +58,7 @@ class Lock(object):
                         self.acquire_timeout)
                 )
 
-    def unlock(self, *exc):
+    def unlock(self):
         with self.get_cursor() as cursor:
             cursor.execute("SELECT RELEASE_LOCK(%s)", (self.name,))
             result = cursor.fetchone()[0]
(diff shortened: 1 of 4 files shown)


adamchainz:
Please add HISTORY note and Docs

adamchainz:
And retitle commit as "Add lock and unlock methods to Lock/TableLock classes"

django_mysql/locks.py
```diff
@@ -52,7 +58,7 @@ def __enter__(self):
                         self.acquire_timeout)
                 )
 
-    def __exit__(self, a, b, c):
+    def unlock(self, *exc):
```
adamchainz: `exc` is unused, please don't pass it in

django_mysql/locks.py
```diff
@@ -144,9 +150,9 @@ def lock(self):
                 locks.append("{} WRITE".format(qn(name)))
             cursor.execute("LOCK TABLES {}".format(", ".join(locks)))
 
-    def unlock(self, exc_type, exc_value, traceback):
+    def unlock(self, *exc):
```
adamchainz: This won't work if one doesn't pass in `exc_type, exc_value, traceback`, please refactor to:

`def unlock(self, exc_type=None, exc_value=None, traceback=None):`

and leave the `_atomic.__exit__` call unchanged

adamchainz:
Add add tests

adamchainz:
Merged in 24a19ccb5f3464886f8de7bf2346d902df94b248, thanks! I thought `acquire` and `release` would be better names since they mirror `threading.Lock`'s methods.

PR #254, opened 2016-03-02
Add 'If' Database Function

### LINKED ISSUE #249: Support SQL If()
Summary: MySQL has a handy `If()` statement, it's shorter than doing the whole `Case(When(` malarkey, we should support it.

adamchainz:
Summary: Fixes #249 .

Checklist:

- [x] Docs updated
- [x] Line added to HISTORY.rst
- [x] Added tests

Test Plan: Lots of tests


adamchainz:
On second thoughts, I don't like the automatic `Value()` wrapping; referring to columns will be more usual than passing a string to the DB.

PR #284, opened 2016-05-02
Fix index hint for nested queries

VladimirKuzmin:
Summary:
Using index hint on a query where indexed table represented only in subquery (e.g. if you use Paginator, some tables might be ommited in count() query) causes syntax error (the hint added before table alias).

```python
class Post(Model):
    author = ForeignKey('User')
    text = TextField()
    published = DateTimeField()

class TaggedPost(Post):
    tags = ManyToManyField('Tag')

# use subquery instead of distinct() for optimization purposes; distinct() causes filesort
post_ids = TaggedPost.objects.filter(tags__in=tags_list).values('pk')
posts = TaggedPost.objects.filter(pk__in=post_ids).use_index('published_idx', table_name='post')
# next line generate invalid query
count = posts.count()
```

Fix rely on weak assumption that django uses aliases for tables in nested queries and these aliases match `[A-Z]+\d+`. From now on `use_index()` doesn't apply hint to any aliased tables.

Checklist:
- [ ] Line added to HISTORY.rst
- [ ] Self added to AUTHORS.rst (or you can opt out)

Test Plan:
Tested in my project with and without nested queries. Added test case.

### DIFF
--- a/django_mysql/rewrite_query.py
+++ b/django_mysql/rewrite_query.py
@@ -167,7 +167,7 @@ def modify_sql(sql, add_comments, add_hints, add_index_hints):
 table_spec_re_template = r'''
     \b(?P<operator>FROM|JOIN)
     \s+
-    {table_name}
+    {table_name}(?!\s+[A-Z]+\d+)
     \s+
 '''
 
--- a/tests/testapp/test_rewrite_query.py
+++ b/tests/testapp/test_rewrite_query.py
@@ -310,6 +310,26 @@ class RewriteQueryTests(TestCase):
             "WHERE (1) ORDER BY col_a"
         )
 
+    def test_index_hint_not_applied_to_nested(self):
+        assert (
+            rewrite_query(
+                "SELECT COUNT(*) FROM `sometable` "
+                "WHERE id IN ("
+                "  SELECT DISTINCT U1.id FROM `sometable` U0"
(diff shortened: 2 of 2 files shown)


adamchainz:
Hey, this looks awesome. Could you please add a line to `HISTORY.rst` explaining that the fix was added, and add yourself to AUTHORS.rst if you like!

I'm going to have a bit of a test before I merge this too.

adamchainz:
Hey, just had a closer look. This needs more testing. Specifically I'd like to see:

- [ ] a test with the query hints in the inner query
- [ ] a test with hints in the inner and outer queries
- [ ] a test with the *same table* in the inner and outer queries, each with hints, and the query hint from the inner query applying to the inner one, the hint from the outer one applying to the outer one (I suspect this won't work :( )
- [ ] tests at the `QuerySet` level too - probably versions of all the above test cases. It's fine if this works for the SQL that Django generates for subqueries atm, but it might not be in the future, and we need to be protected against that. Such tests are in `test_models.py::QueryHintTests`

VladimirKuzmin:
Hey!
I'll add more tests.

VladimirKuzmin:
Hey, Adam!
You're right this solution won't work for any inner queries.
I considered better solution for this problem. Have a look at PR https://github.com/adamchainz/django-mysql/pull/286.

PR #285, opened 2016-05-05
Upgrade requirements

adamchainz:


adamchainz:
Test failures are all Django master, will fix those in another PR

PR #286, opened 2016-05-08
Index hint for nested queries (bound to alias)

VladimirKuzmin:
Summary:

Index hint should remember which table it was applied.
Solution:
- on each aliases changing in query also change alias in index hint
- at rewriting stage modify SQL according the alias

Checklist:
- [x] Docs updated, or N/A
- [x] Added extra items to checklist as applicable

Test Plan:
- added tests for nested queries up to 3 level generated by Django ORM
- added test for raw SQL query

Known issue:
For nested queries like this (without any `JOIN`): 

``` python
SomeModel.objects.filter(pk__in=SomeModel.objects.all())
```

Django will not assign alias to any table, so index hint will be always applied to outer query.
But it looks like queries like this are useless.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -13,3 +13,4 @@ Contributors
 * Łukasz Więcek <https://github.com/lwiecek>
 * Marco Santamaria <https://github.com/marco-santamaria>
 * Michael Aquilina <https://github.com/MichaelAquilina>
+* Vladimir Kuzmin <https://github.com/VladimirKuzmin>
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,8 @@ Pending
 
 * New release notes here
 * Fixed some features to work when there are non-MySQL databases configured
+* Fixed syntax error when index hint used on an aliased table
+* Index hints apply to the correct table entry in nested queries
 
 1.0.8 (2016-04-08)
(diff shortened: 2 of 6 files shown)


adamchainz:
This is nice and looks good. A quick glance doesn't reveal anything problematic, but I'll give it a bit more testing during the week!

thebarbershop:
Can someone share the status of this patch? Is the issue addressed by some other PR, or is it just left unfixed for some reason?

adamchainz:
I've been meaning to review this but it's quite a big chunk of stuff to understand at once (and after this time, rebase). And I've never had the issue myself, and I guess I've always had higher impact things to work on.

If you want to rebase it that would be a helpful start; even more helpful would be running it in your app (dev, or staigng, or even prod) and seeing if you encounter any issues not addressed by tests.

thebarbershop:
Thanks for the update, Adam!

I'm currently working to upgrade a project from ancient times (Django 1.9 + A forked django-mysql 1.0.9 with this PR applied) so I can't contribute to the upstream at this point. But I'll be happy to return to this issue when I can.

adamchainz:
I'm calling it quits on this one. I don't have any time or motivation to return to it, I'm afraid, and no one else has stepped forward.

Issue #222, opened 2016-02-02
Remove Django 1.7 support


adamchainz:
As per Django's [Supported Versions Grid](https://www.djangoproject.com/download/), 1.7 support ended December 2015, so we can remove it too.

PR #292, opened 2016-05-20
Fix build for Django 1.10 changes

adamchainz:
- [x] Stop building Django master, fix on the 1.10 branch since it's alpha now
- [x] Fix for `get_prep_lookup` / `get_db_prep_lookup` obsoletion in django/django#6499
- [x] Fix `call_command('migrate')` requiring `interactive=False`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,7 +68,7 @@ Requirements
 Tested with all combinations of:
 
 * Python: 2.7, 3.4, 3.5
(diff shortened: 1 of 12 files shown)


adamchainz:
The changes aren't compatible with Django 1.7... can't decide whether or not to do #222 before this...

adamchainz:
Woops still need to fix `JSONField`

Issue #345, opened 2017-03-27
Numpy NaN is not supported by JSON.


arnau126:
Summary: NaN is not in the JSON specification, so when you try to save a django model which contains a numpy nan, it crashes with a misleading error message:

> InterfaceError: (-1, 'error totally whack')

I wonder if it is possible for django_mysql to also serialize NaN (as json.dumps does).

If not, at least raise an easier to understand message error.
This could be achieve by calling json.dumps with allow_nan=False.
Doing this, the raised error would be:

> ValueError: Out of range float values are not JSON compliant

(and is raised by json.dumps function, instead of by the DB)

Django Version: *Django 1.10.6*

Database and version used: *MySQL 5.7*

Version: *Django-MySQL 1.1.0*

adamchainz:
Is it `numpy.nan` or `float('nan')` you're having a problem with?

Anyway seems like we could just pass `allow_nan=False` to `json.dumps` in the field here: https://github.com/adamchainz/django-mysql/blob/master/django_mysql/models/fields/json.py#L102

arnau126:
@adamchainz 

> Is it numpy.nan or float('nan') you're having a problem with?

Both, since json.dumps serializes both in the same way.

> Anyway seems like we could just pass allow_nan=False to json.dumps in the field here: https://github.com/adamchainz/django-mysql/blob/master/django_mysql/models/fields/json.py#L102

Yes, this is what I meant.

adamchainz:
Ok PR welcome!

arnau126:
PR: #346

PR #346, opened 2017-03-28
Disallow NaN in JSONField because MySQL doesn't support it

arnau126:
Summary: Call json.dumps with allow_nan=False, to get a ValueError instead of a misleading DB error, in case of 'value' containing NaN.

Issue: #345

Test Plan: Test added. All tests pass.

### DIFF
--- a/tests/testapp/test_jsonfield.py
+++ b/tests/testapp/test_jsonfield.py
@@ -66,6 +66,15 @@ class TestSaveLoad(JSONFieldTestCase):
         m = JSONModel.objects.get()
         assert m.attrs == [chars]
 
+    def test_nan_raises_valueerror(self):
+        m = JSONModel(attrs=float('nan'))
+        try:
+            m.save()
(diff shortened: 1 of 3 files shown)


tests/testapp/test_jsonfield.py
```diff
@@ -66,6 +66,15 @@ def test_control_characters(self):
...
         assert m.attrs == [chars]
 
+    def test_nan_raises_valueerror(self):
+        m = JSONModel(attrs=float('nan'))
+        try:
+            m.save()
+        except ValueError:
+            pass
+        else:
+            assert False
```
adamchainz: Use `with pytest.raises(ValueError):` rather than the `try`/`except`

arnau126:
Requested changes done.

adamchainz:
Released as version 1.1.1 https://pypi.python.org/pypi/django-mysql

Issue #348, opened 2017-04-03
JSONField form: don't dump if value is already a string. 


arnau126:
I have a Django model with a JSONField, and I want to create an instance of it though the admin site. 
In the JSONField I write the following:
```
[
    {"name": "foo"},
    {"name": "bar"}, 
] 
```
As expected, it's an invalid JSON (because of the trailing comma) and a ValidationError is raised.

What is unexpected is that when the form is re-rendered the content of the json field become:
`
" \r\n    [{\"name\": \"foo\", \r\n    {\"name\": \"bar\"}, \r\n]"
`
which is not json valid at all. 

This happens because the `prepare_value` function of the JSONField form is dumping the content of the field, even when it's already a string. 
https://github.com/adamchainz/django-mysql/blob/master/django_mysql/forms.py#L250

Django Version: *Django 1.10.6*

Database and version used: *MySQL 5.7*

Version: *Django-MySQL 1.1.1*

arnau126:
I submitted a PR #349.

PR #349, opened 2017-04-03
Don't dump value if its already a string in JSONField.prepare_value.

arnau126:
Make that `prepare_value` function of JSONField form doesn't dumb value if it's already a string.
It happens when submitting an invalid JSON into a form, that results into a "more invalid" json.

Example in #348.

Checklist:

- [ ] Docs updated, or N/A

Test Plan: All current tests pass, and extra test added.

### DIFF
--- a/django_mysql/forms.py
+++ b/django_mysql/forms.py
@@ -247,4 +247,6 @@ class JSONField(forms.CharField):
             )
 
     def prepare_value(self, value):
-        return json.dumps(value)
+        if not isinstance(value, six.string_types):
+            return json.dumps(value)
+        return value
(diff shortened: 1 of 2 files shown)


adamchainz:
Doesn't this break storing strings in JSON, which should serialize to `"string"` ? I think there's some other thing we should be doing for form processing, can't remember what.

adamchainz:
Actually no, storing strings is broken. Will update the docs accordingly. Your PR seems sane.

arnau126:
I've submitted another PR (which invalidates this one) that solve this issue and also expects to fix the feature of storing strings in json: #350

PR #350, opened 2017-04-04
Storing strings in JSONField.

arnau126:
Be able to store strings in JSONField (this feature was broken).

Also copy code from 'django.contrib.postgres.forms.jsonb.JSONField' in order to fix some issues related with redisplaying wrong user input.

This PR is motivated by: #349 (which now it is not longer necessary).

Checklist:

Test Plan:
- Create some test about serializing strings.
- Copy two tests from ' django.tests.postgres_tests.test_json.py'.
- All current and new tests pass.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,8 @@ Pending
 -------
 
 * (Insert new release notes below this line)
+* Fixed ``JSONField`` to allow storing strings as top-level json objects.
+* Fixed ``JSONField`` so that it doesn't overquote inputs when redisplaying the form due to invalid user input.
 
 1.1.1 (2017-03-28)
 ------------------
--- a/django_mysql/forms.py
+++ b/django_mysql/forms.py
@@ -240,8 +240,6 @@ class JSONField(forms.CharField):
     widget = forms.Textarea
 
     def to_python(self, value):
-        if self.disabled:
-            return value
         if value in self.empty_values:
             return None
         elif isinstance(value, (list, dict, int, float, JSONString)):
@@ -260,8 +258,6 @@ class JSONField(forms.CharField):
             return converted
 
     def bound_data(self, data, initial):
-        if self.disabled:
-            return initial
         try:
             return json.loads(data)
         except ValueError:
--- a/django_mysql/forms.py
+++ b/django_mysql/forms.py
@@ -225,26 +225,49 @@ class SimpleSetField(forms.CharField):
             raise ValidationError(errors)
 
 
+class InvalidJSONInput(six.text_type):
+    pass
+
+
+class JSONString(six.text_type):
+    pass
+
+
 class JSONField(forms.CharField):
     default_error_messages = {
         'invalid': _("'%(value)s' value must be valid JSON."),
     }
-
-    def __init__(self, **kwargs):
-        kwargs.setdefault('widget', forms.Textarea)
-        super(JSONField, self).__init__(**kwargs)
+    widget = forms.Textarea
 
     def to_python(self, value):
+        if self.disabled:
+            return value
         if value in self.empty_values:
             return None
+        elif isinstance(value, (list, dict, int, float, JSONString)):
+            return value
         try:
-            return json.loads(value)
+            converted = json.loads(value)
         except ValueError:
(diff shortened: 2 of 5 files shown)


arnau126:
TODO: tests don't pass in `django1.8` because in this version `Field` doesn't have the `disabled` attribute, yet.

arnau126:
any thoughts, @adamchainz ?

sergiomb2:
@arnau126 	Merge branch 'master' into string_jsons 

instead merge , you use rebase , git rebase upstream/master && git push --force (I think you need force the push)

adamchainz:
@arnau126 don't do anything I'll rebase the PR

I have been a bit busy recently but will look at this soon.

arnau126:
@sergiomb2 Thanks for the advice.

@adamchainz Okay. Let me know if any change is needed.

arnau126:
@adamchainz Sorry for the rush, but have you been able to take a look at this?
Besides, when do you expect to release? I would like to use JSONSet in the project I'm currently working on.

adamchainz:
Looking at it now finally.

I am going to merge this as two commits, one for the form field and one for the model field. I think the model field fix could break some people's applications if they've been doing e.g. `model.attrs = json.dumps({'a': 'dict'})`, so I think it should go in a "breaking change" release, just in case.

HISTORY.rst
```diff
@@ -7,6 +7,8 @@ Pending
 -------
 
 * (Insert new release notes below this line)
+* Fixed ``JSONField`` to allow storing strings as top-level json objects.
+* Fixed ``JSONField`` so that it doesn't overquote inputs when redisplaying the form due to invalid user input.
```
adamchainz: this file is wrapped at 80 chars

django_mysql/forms.py
```diff
@@ -240,8 +240,6 @@ class JSONField(forms.CharField):
     widget = forms.Textarea
 
     def to_python(self, value):
-        if self.disabled:
```
adamchainz: removing this logic isn't the right approach, we want to work with disabled on Django 1.9+. instead we should do:

```
if getattr(self, 'disabled', False):
```

or

```
if django.VERSION[:2] >= (1, 9) and self.disabled:
```

Allowing the code to work on both versions. And add a test too, to get coverage of the lines.

tests/testapp/test_forms.py
```diff
@@ -306,4 +307,37 @@ def test_invalid(self):
...
+        When displaying a bound form (typically due to invalid input), the form
+        should not overquote JSONField inputs.
+        """
+        class JsonForm(forms.Form):
+            name = forms.CharField(max_length=2)
+            jfield = JSONField()
+
+        # JSONField input is fine, name is too long
+        form = JsonForm({'name': 'xyz', 'jfield': '["foo"]'})
+        self.assertIn('[&quot;foo&quot;]</textarea>', form.as_p())
```
adamchainz: our tests are on pytest and use plain `assert` everywhere, so this shoudl be `assert '<snip>' in form.as_p()` etc.

adamchainz:
Merged form field change in 9930b2009c72cea7d2bd15c474e0edd209e5138a with some touch-up in 556d8987845fb3a11f73a2a86855fd311126967f

arnau126:
Okay!
And thanks for all your remarks.

kyteague:
FYI, the bug that this PR addresses caused us two production outages because bad JSON data was saved back to the database (our application was expecting a JSON list not a string). A quick merge would be appreciated, thanks!

arnau126:
This bug is already merged and released in 1.2.0:
https://github.com/adamchainz/django-mysql/blob/master/HISTORY.rst#120-2017-05-14

adamchainz said:
> I am going to merge this as two commits, one for the form field and one for the model field.

> Merged the form field stuff in 2f03d96 , fixed up in 504fd9c. Thanks!

adamchainz:
I think @kyteague wants the model field fix. I'm going to merge it today

adamchainz:
Merged the model field fix in 961b9f5c817a599e3a6f4a033a63d7e5cf326213, touched up in 12616152e5129559daacba8957cb738319eee586. Thanks!

Issue #347, opened 2017-03-29
Add JSON Create & Modify MySQL functions 


avelis:
Summary: To have the ability to use MySQL JSON functions that can create and modify data as Django ORM expressions and/or functions.

e.g.
```python
shop_item.attrs = JSONSet(F('attrs'), {'$.smelliness': 10, '$.color': 'magenta'})
```
This is one interpretation of what I am essentially getting at. It could be possible that the JSONSet class could also augment the keys to prepend the `$.` string.

I could attempt a PR for this but would like some direction on this.

adamchainz:
Yes I agree they would be good to add. I think I'm against the key-augmentation thing, because the keys in that dict would be json paths and could feature complex expressions like `$.colours[0][1]`.

avelis:
@adamchainz ok, that makes sense.

avelis:
```

adamchainz:
would prefer it in a pull-request with tests :)

Also the `len` thing would be better done at the top with a check on `pairs` being a dict of at least one key.

avelis:
@adamchainz I can make a PR. I wrote this up in the late hours of the night and felt it needed a second eye before making a PR.

adamchainz:
PR = second set of eyes at code review 😉

avelis:
ha! Fair enough.

PR #351, opened 2017-04-05
Add JSONSet function class

avelis:
Addresses #347

Class will allow users at the ORM layer to specify the JSON_SET SQL
function in mysql when modifying JSON data.

### DIFF
--- a/django_mysql/models/functions.py
+++ b/django_mysql/models/functions.py
@@ -266,9 +266,9 @@ class JSONSet(Func):
     def __init__(self, expression, pairs):
         from django_mysql.models.fields import JSONField
 
-        if len(pairs) == 0:
+        if not len(pairs):
             raise ValueError(
-                'Pairs requires one key/value set defined'
+                '"pairs" requires at least one entry'
             )
 
         exprs = [expression]
--- a/django_mysql/models/functions.py
+++ b/django_mysql/models/functions.py
@@ -267,7 +267,9 @@ class JSONSet(Func):
         from django_mysql.models.fields import JSONField
 
         if len(pairs) == 0:
-            raise ValueError('Pairs argument requires one key/value set defined')
+            raise ValueError(
+                'Pairs requires one key/value set defined'
+            )
 
         exprs = [expression]
 
--- a/django_mysql/models/functions.py
+++ b/django_mysql/models/functions.py
@@ -266,12 +266,12 @@ class JSONSet(Func):
     def __init__(self, expression, pairs):
         from django_mysql.models.fields import JSONField
 
-        if isinstance(pairs, dict) and len(pairs.keys()) == 0:
+        if len(pairs) == 0:
             raise ValueError('Pairs argument requires one key/value set defined')
 
         exprs = [expression]
 
-        for path, value in pairs.iteritems():
+        for path, value in pairs.items():
             if not hasattr(path, 'resolve_expression'):
                 path = Value(path)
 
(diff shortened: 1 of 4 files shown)


avelis:
@adamchainz Will write a test soon. Wanted to start the PR either way.

django_mysql/models/functions.py
```diff
@@ -260,6 +260,31 @@ def __init__(self, expression, path=None, **extra):
         super(JSONLength, self).__init__(*exprs, output_field=output_field)
 
 
+class JSONSet(Func):
+    function = 'JSON_SET'
+
+    def __init__(self, expression, pairs):
+        from django_mysql.models.fields import JSONField
+
+        if isinstance(pairs, dict) and len(pairs.keys()) == 0:
```
adamchainz: 1. Remove the `isinstance` - firstly, the code currently doesn't block non-dicts, and secondly, we can just allow duck-typing of dict-like objects, no need for strict typing
2. Use `len(pairs)`, no need to extract `keys`

django_mysql/models/functions.py
```diff
@@ -260,6 +260,31 @@ def __init__(self, expression, path=None, **extra):
...
+            raise ValueError('Pairs argument requires one key/value set defined')
+
+        exprs = [expression]
+
+        for path, value in pairs.iteritems():
```
adamchainz: `iteritems` doesn't exist on Python 3, just use `items` since we expect the dict to be relatively small

django_mysql/models/functions.py
avelis: Yeah, I realize my haste made waste on the `isinstance` check. Sigh.

django_mysql/models/functions.py
```diff
@@ -260,6 +260,33 @@ def __init__(self, expression, path=None, **extra):
...
+class JSONSet(Func):
+    function = 'JSON_SET'
+
+    def __init__(self, expression, pairs):
+        from django_mysql.models.fields import JSONField
+
+        if len(pairs) == 0:
+            raise ValueError(
+                'Pairs requires one key/value set defined'
+            )
```
adamchainz: `if not len(pairs)` is even shorter :)

Also I'd write `'paris' requires at least one entry`

avelis: I considered that type of if check. Wasn't sure though. But I can do that as well.

Also, I presume you meant `'pairs'` and not `'paris'`?

avelis:
@adamchainz Is there a specific way to add to the docs?

adamchainz:
Yes that's the right place. Put it with the other json funcs in alphabetic order, i.e. after `JSONLength`

avelis:
@adamchainz Docs added.

adamchainz:
Merged in 4fe3a266e757bf7f71189a7e6c98e95164beea96, thanks!

adamchainz:
Btw I made some small changes before merging:

1. Wrapped text at 80 chars in database_functions.rst - this is the format for all the docs, you ignored it when writing there
2. Changed the 'pairs' dictionary to 'data' - 'pairs' is a funny name for a dict to me, it sounds like a list of tuples rather than a dictionary.
3. Changed the example and test to not use `F()` to wrap the first argument `expression` because it's not necessary

avelis:
@adamchainz I appreciate the merge.

I did not know about the 80 characters text wrap for the docs. Is that a convention with the open source community in general?

Not being satirical at all. I am asking for personal knowledge growth. Thanks.

adamchainz:
Django and most of its community projects have a soft 80 / hard 120 wrap rule. But in general, if you edit a file, look at the lines around for context - even without editor rulers on, it should be clear the lines are being wrapped at some width 😉  And in more general - if you're going to edit a file in a project, respect the conventions in that file and project, even if it's not your style.

adamchainz:
Released in version 1.2.0

PR #352, opened 2017-04-18
Use custom version of six included with Django

flanaman:
Use django.utils.six everywhere instead of the version from PyPi. Ie, changed `import six` to `from django.utils import six`. Most imports for `six` used the django version already, but two did not. `six` is not listed in the pip install requirements (though they are in the test environment requirements), so trying to use, say, EnumField fails if `six` is not installed.

Checklist:

- [x] Docs updated, or N/A
- [x] Added extra items to checklist as applicable

Test Plan: tox passed for python 3.6 and django 1.8, 1.1.9, 1.10. (tox didn't run for other python versions due to unrelated local python installation peculiarities.)

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,7 +7,8 @@ Pending
 -------
 
 * (Insert new release notes below this line)
-* Use django.utils.six everywhere
+* Use django.utils.six everywhere, fixes usage of ``EnumField`` and field
+  lookups when ``six`` is not installed (from PyPI)
 
 1.1.1 (2017-03-28)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,7 @@ Pending
 -------
 
 * (Insert new release notes below this line)
+* Use django.utils.six everywhere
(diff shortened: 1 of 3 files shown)


adamchainz:
Thanks. Can you add to the HISTORY note that this fixes usage when `six` is not installed? One thing is though, we might have to flip everything to standalone `six` from PyPI for Django 2.0 or beyond, as Django will stop vendoring it as 2.0+ don't support Python 2.

flanaman:
> we might have to flip everything to standalone six from PyPI for Django 2.0 or beyond, as Django will stop vendoring it as 2.0+ don't support Python 2.

I don't think it would be a big deal to put `six` into the setup.py requirements for this package, just figured `django.utils.six` was simpler for now. The tests should pick up the absence of `django.utils.six` immediately when that happens.

adamchainz:
Merged in d0deca9414d50b7b83ca63a8bb84289ff1716414, thanks!

adamchainz:
Released in version 1.2.0

PR #361, opened 2017-05-12
Improve JSONSet, JSONReplace, JSONInsert by allowing dicts and lists.

arnau126:
Current `JSONSet` function only allows to set values like `ints`, `floats` and `str`.

This PR improves/fixes this function by also allowing to set `dicts` and `lists`.

For example:
```
obj = JSONModel.objects.create(attrs={})
obj.attrs = JSONSet('attrs', {'$.data': {'foo': 'bar', 'nums': [11, 38, 42]}})
obj.save()
```
Which will result in:

```
obj.refresh_from_db()
assert obj.attrs['data'] == {'foo': 'bar', 'nums': [11, 38, 42]}
```

The same for `JSONInsert` and `JSONReplace`.

Checklist:

- [x] Docs updated, or N/A

Test Plan:
- For each class ( `JSONSet`, `JSONInsert` and `JSONReplace`) add one test for `dicts` and one for `lists`.
- Add a more complex test (multi-level nested dicts and lists) just for `JSONSet`. I assume that it also works for the other classes as all of them inherits from `BaseJSONModifyFunc`.
- All current and new tests pass.

### DIFF
--- a/django_mysql/models/functions.py
+++ b/django_mysql/models/functions.py
@@ -260,6 +260,17 @@ class JSONLength(Func):
         super(JSONLength, self).__init__(*exprs, output_field=output_field)
 
 
+class JSONValue(Func):
+    function = 'CAST'
+    template = '%(function)s(%(expressions)s as json)'
+    arity = 1
+
+    def __init__(self, expression):
+        import json
+        value = json.dumps(expression, allow_nan=False)
+        super(JSONValue, self).__init__(Value(value))
+
+
 class BaseJSONModifyFunc(Func):
     def __init__(self, expression, data):
         from django_mysql.models.fields import JSONField
@@ -276,7 +287,7 @@ class BaseJSONModifyFunc(Func):
             exprs.append(path)
 
             if not hasattr(value, 'resolve_expression'):
-                value = Value(value)
+                value = JSONValue(value)
 
             exprs.append(value)
 
--- a/tests/testapp/test_functions.py
+++ b/tests/testapp/test_functions.py
@@ -432,6 +432,28 @@ class JSONFunctionTests(JSONFieldTestCase):
         assert self.obj.attrs['int'] == 88
         assert self.obj.attrs['int2'] == 102
 
+    def test_json_insert_dict(self):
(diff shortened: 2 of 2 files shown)


arnau126:
I didn't document in HISTORY.rst because `JSONSet` is not released yet.
So I thought that the feature of this PR is implicit in:

> - Added JSONInsert, JSONReplace and JSONSet database functions that wraps the respective JSON-modifying functions from MySQL 5.7.

(https://github.com/adamchainz/django-mysql/blob/master/HISTORY.rst#pending)

arnau126:
I amended my commit with the requested changes.
Let me know if this is what you meant.

adamchainz:
Good point on the HISTORY note 👍

django_mysql/models/functions.py
```diff
@@ -260,6 +260,17 @@ def __init__(self, expression, path=None, **extra):
         super(JSONLength, self).__init__(*exprs, output_field=output_field)
 
 
+class JSONValue(Func):
+    function = 'CAST'
+    template = '%(function)s(%(expressions)s as json)'
+    arity = 1
+
+    def __init__(self, expression):
+        import json
```
adamchainz: inner import not required, do it at the module level to make the `__init__` faster

django_mysql/models/functions.py
```diff
@@ -260,6 +260,17 @@ def __init__(self, expression, path=None, **extra):
         super(JSONLength, self).__init__(*exprs, output_field=output_field)
 
 
+class JSONValue(Func):
+    function = 'CAST'
+    template = '%(function)s(%(expressions)s as json)'
+    arity = 1
```
adamchainz: no need for this as we are defining custom `__init__` so it's impossible to init with wrong arity

adamchainz:
Merged in 9930b2009c72cea7d2bd15c474e0edd209e5138a with some touch-up in 556d8987845fb3a11f73a2a86855fd311126967f, thanks!

adamchainz:
Released in version 1.2.0

arnau126:
@adamchainz Thanks to you!

PR #363, opened 2017-05-28
Django 1.11 compatibility

adamchainz:
Summary: Adds Django 1.11 compatibility

Checklist:

- [x] Docs updated
- [x] Line added to HISTORY.rst

Test Plan: Updated `tox.ini` and some tests that failed

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,7 @@ Pending
 -------
 
 * (Insert new release notes below this line)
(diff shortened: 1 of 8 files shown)


adamchainz:
Really the only problem came from django/django#7587 which stopped there being a nice single method (`iterator()`) in which to inject the found_rows fetching logic, but the workaround with `_result_cache` is not that bad.

PR #368, opened 2017-06-13
Json search

patr1otkz:
Summary:  
Added the ability to search by the key of the object, if the key is a number. Before a number it is necessary to put any letter. Example: json = {'1': '2'},  queryset = Model.objects.filter(json_field__k1='2')
Added MySQL function JSON_SEACR to find the value inside the object. Example:
json = {'key1': 'value1', 'obj': {'key2': 'value2'}, 'key3': 'value3'} 
queryset = Model.objects.filter(json_field__search='value') 
https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html#function_json-search

Checklist:

- [ ] Docs updated, or N/A

Test Plan: Works with the current CRM project

### DIFF
--- a/django_mysql/models/fields/json.py
+++ b/django_mysql/models/fields/json.py
@@ -4,6 +4,7 @@ from __future__ import (
 )
 
 import json
+import re
 
 import django
 from django.core import checks
@@ -15,7 +16,7 @@ from django_mysql import forms
 from django_mysql.models.lookups import (
     JSONContainedBy, JSONContains, JSONExact, JSONGreaterThan,
     JSONGreaterThanOrEqual, JSONHasAnyKeys, JSONHasKey, JSONHasKeys,
-    JSONLessThan, JSONLessThanOrEqual
+    JSONLessThan, JSONLessThanOrEqual, JSONSearch
 )
 from django_mysql.utils import collapse_spaces, connection_is_mariadb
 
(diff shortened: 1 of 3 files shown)


arnau126:
@patr1otkz, there is already a way to to search by the key of the object, if the key is a number:
`queryset = Model.objects.filter(**{'json_field__"1"': '2'})`.

sergiomb2:
Hi, 
I'm looking for something like this , I get [1]

the SQL is simple something like :  
`SELECT * FROM table WHERE json_field COLLATE utf8mb4_unicode_ci LIKE '%ICT%' `

[1]
**NotImplementedError at /themis/api/ci/
Lookup 'icontains' doesn't work with JSONField**

key  | value 
-- | --
Request Method: |	GET
Request URL: 	| http://127.0.0.1:51002/api/table/?search=ICT
Django Version: |	1.11.3
Exception Type: |	NotImplementedError
Exception Value: |	Lookup 'icontains' doesn't work with JSONField
Exception Location:  |	/usr/lib/python3.5/site-packages/django_mysql/models/fields/json.py in get_lookup, line 124
Python Executable: 	| /usr/bin/python3
Python Version: |	3.5.3

adamchainz:
@sergiomb2 that's a completely separate issue to this PR, please don't report it here, make an issue, then I'll respond.

adamchainz:
@arnau126 was right about the keys. To be sure, I'm adding some tests in #381 to check this.

As for your JSON_SEARCH behaviour, it looks useful but a bit too specific. I'll tackle adding a full `JSONSearch` function that can be used to perform any possible search rather than this one specific one.

Thanks for the suggestions though!

PR #408, opened 2017-11-05
Hide deprecation warnings from MySQL 5.7

adamchainz:
MySQL 5.7 emits warnings for some deprecated stuff, which pollute our test output. This stops them warning.


adamchainz:
Didn't work, yet...

PR #412, opened 2017-11-13
Some code improvements from lgtm.com

adamchainz:
Most of its alerts are annoying, but at least these changes make the code more robust.

### DIFF
--- a/django_mysql/cache.py
+++ b/django_mysql/cache.py
@@ -212,6 +212,9 @@ class MySQLCache(BaseDatabaseCache):
         return self._base_set('add', key, value, timeout)
 
     def _base_set(self, mode, key, value, timeout=DEFAULT_TIMEOUT):
+        if mode not in ('set', 'add'):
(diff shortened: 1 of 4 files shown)


adamchainz:
fyi @MichaelAquilina

django_mysql/cache.py
```diff
@@ -224,15 +224,15 @@ def _base_set(self, mode, key, value, timeout=DEFAULT_TIMEOUT):
             if mode == 'set':
                 query = self._set_query
                 params = (key, value, value_type, exp)
-            elif mode == 'add':
+            else:  # mode = 'add'
```
MichaelAquilina: this one seems odd 🤔 wouldnt something more appropriate be

```
elif mode == 'add':
   # stuff
else:
  raise ValueError('invalid mode specified')
```

adamchainz: Yeah you're right, even though it's internal only and will be a pain to get coverage on

PR #419, opened 2017-12-05
Upgrade requirements

adamchainz:


adamchainz:
Broken because of pytest-dev/pytest#2956

PR #426, opened 2018-02-08
ensure_ascii=False param for json.dumps

Yolley:
JSON may contain non-ASCII characters, and, if we edit it in django admin, in form these characters will appear in non-readable form, for example:
["\u0442\u0435\u0441\u0442"]
Adding "ensure_ascii=False" param will fix this.

Checklist:

- [ ] Docs updated, or N/A

Test Plan: tested only on my project

### DIFF
--- a/django_mysql/forms.py
+++ b/django_mysql/forms.py
@@ -271,4 +271,4 @@ class JSONField(forms.CharField):
     def prepare_value(self, value):
         if isinstance(value, InvalidJSONInput):
             return value
-        return json.dumps(value)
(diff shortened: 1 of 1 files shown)


adamchainz:
Is this something the model field should be doing too?

Please also complete the checklist rather than delete it:

* Add tests
* Add note in HISTORY

adamchainz:
@Yolley do you still want to work onthis?

adamchainz:
Closing for now, if you continue having issues please come back to me

PR #445, opened 2018-04-11
Fix crashes for DynamicField instances without spec

hyzyla:
Changed None to empty dict for empty constructor parameter in dynamic field

Summary: 
"None" isn't a good choice for spec, on my opinions, because during migrations on several projects I have always get Attribute error in validators and other functions that work with "spec" as with dictionary and doesn't check if it dictionary or not.

Checklist:

- [ ] Docs updated, or N/A

Test Plan: Manually

### DIFF
--- a/django_mysql/models/fields/dynamic.py
+++ b/django_mysql/models/fields/dynamic.py
@@ -35,7 +35,7 @@ class DynamicField(Field):
             kwargs['default'] = dict
         if 'blank' not in kwargs:
             kwargs['blank'] = True
-        self.spec = kwargs.pop('spec', None)
+        self.spec = kwargs.pop('spec', {})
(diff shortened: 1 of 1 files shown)


adamchainz:
This is a good start but there are a bunch of other places that would need updating to do this, plus I'd need tests taht fail before and pass afterwards. Also I'd be worried that historical migrations in real world projects would break? It might be better to change all the code to work with `spec=None` properly.

Could you provide more details about the problems you've encountered? It would be useful! Thanks!

adamchainz:
@hyzyla I've looked into this more, I think I've found all the cases where `spec=None` is causing problem and I've added tests, I'll merge your PR here so you get the contribution point and add my stuff in a follow-up PR.

adamchainz:
Fix released in version 2.2.2 on PyPI

PR #467, opened 2018-05-26
Add custom encoder/decoder support in JSONField

### LINKED ISSUE #461: Extend JSONField to use custom encoder/decoder
Summary: How about letting user choose JSON encoder/decoder to use for `JSONField`?

I'm thinking about whether it is feasible and preferable for user to be able to supply a custom JSON encoder (such as `DjangoJSONEncoder`) on `JSONField` initialization like this:

```python
class JSONField(Field):
    encoder = json.JSONEncoder
    decoder = json.JSONDecoder

    def __init__(self, encoder=None, decoder=None, *args, **kwargs):
        self.encoder = encoder or self.encoder
        self.decoder = decoder or self.decoder
        # ...
```

In this way, it's possible to reduce duplicated code on upper layers that has to coerce different
types of values to/from JSON-serializable form.

I'm not sure whether it will potentially break anything cooperating with current `JSONField` such as
JSON functions, as I currently don't have time digging into the code. So I'll just put my two cents here (sorry).

adamchainz:
Fixes #461.

### DIFF
--- a/django_mysql/models/fields/json.py
+++ b/django_mysql/models/fields/json.py
@@ -30,12 +30,8 @@ class JSONField(Field):
     def __init__(self, *args, **kwargs):
         if 'default' not in kwargs:
             kwargs['default'] = dict
-        self.json_encoder = kwargs.pop('encoder', None)
-        if self.json_encoder is None:
-            self.json_encoder = self._default_json_encoder
-        self.json_decoder = kwargs.pop('decoder', None)
-        if self.json_decoder is None:
-            self.json_decoder = self._default_json_decoder
+        self.json_encoder = kwargs.pop('encoder', self._default_json_encoder)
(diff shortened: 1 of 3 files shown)


django_mysql/models/fields/json.py
```diff
@@ -23,16 +23,26 @@
...
+
+    _default_json_encoder = json.JSONEncoder(allow_nan=False)
+    _default_json_decoder = json.JSONDecoder(strict=False)
+
     def __init__(self, *args, **kwargs):
         if 'default' not in kwargs:
             kwargs['default'] = dict
+        self.json_encoder = kwargs.pop('encoder', None)
+        if self.json_encoder is None:
+            self.json_encoder = self._default_json_encoder
```
adamchainz: inline in `pop`, and below

PR #470, opened 2018-05-26
Add touch feature to mysql cache.

### LINKED ISSUE #457: Add cache.touch() method
Coming in Django 2.1 as per django/django#7610, we should support it in MySQLCache

Normos:
Fixes #457


adamchainz:
Version 2.3.0 out with this feature, thanks! https://pypi.org/project/django-mysql/

PR #473, opened 2018-05-26
WIP fix query rewriting for table aliases

### LINKED ISSUE #464: Rewrite query ignores Table alias
Summary:
Rewriting query for index suggestions, ignore table alias causing the generated query to augment the actual query at the wrong place, which causes it to generate invalid queries.

`SELECT TABLE_ALIAS.FIELD1 FROM TABLE_NAME TABLE_ALIAS WHERE ( TABLE_ALIAS.FIELD2=Value AND (/*QueryRewrite':index=`TABLE_NAME` IGNORE `INDEX_TO_IGNORE`*/1))`

Gets rewritten to:

`SELECT TABLE_ALIAS.FIELD1 FROM TABLE_NAME IGNORE INDEX (INDEX_TO_IGNORE) TABLE_ALIAS WHERE ( TABLE_ALIAS.FIELD2=Value AND (1))`

Rather than

`SELECT TABLE_ALIAS.FIELD1 FROM TABLE_NAME TABLE_ALIAS IGNORE INDEX (INDEX_TO_IGNORE) WHERE ( TABLE_ALIAS.FIELD2=Value AND (1))`

Basically the regex that rewrite uses ignore any table alias. Currently if I pass a querySet to another querySet, django creates an Alias for the table, which breaks the rewrite translation.

The bug lies in the regex `table_spec_re_template` at https://github.com/adamchainz/django-mysql/blob/master/django_mysql/rewrite_query.py#L169.  Which the rewrite uses to describe the table definition.

Django Version: 2.0.5

Database and version used: MySQL 5.7

Version: *e.g. Django-MySQL 1.0.1*

adamchainz:
Fixes #464

### DIFF
--- a/django_mysql/rewrite_query.py
+++ b/django_mysql/rewrite_query.py
@@ -169,13 +169,17 @@ def modify_sql(sql, add_comments, add_hints, add_index_hints):
 table_spec_re_template = r'''
     \b(?P<operator>FROM|JOIN)
     \s+
-    {table_name}
+    (?P<table>
+        {table_name}
+        (\ [^ ]+)?  # alias/WHERE
+    )
     \s+
+    WHERE
 '''
 
 replacement_template = (
-    r'\g<operator> {table_name} '
-    r'{rule} INDEX {for_section}({index_names}) '
+    r'\g<operator> \g<table> '
+    r'{rule} INDEX {for_section}({index_names}) WHERE'
 )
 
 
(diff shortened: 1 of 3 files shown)


adamchainz:
This doesn't currently work. Parsing the alias out is actually quite tricky, we might be able to make it work if we restrict the universe of supported aliases to just those that Django generates though (`U0`, `U1`, etc.)

adamchainz:
Abandoning due to age.

PR #475, opened 2018-05-26
Test with MariaDB 10.3

### LINKED ISSUE #471: Test on MariaDB 10.3
Since it's out now, we should test against it

Normos:
Fixes #471.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,11 @@ env:
     - DB=mysql DB_VERSION=5.6
     - DB=mysql DB_VERSION=5.7
     # Can't get this to work, as above
-    # - DB=mariadb DB_VERSION=5.5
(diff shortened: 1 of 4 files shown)


Normos:
MariaDB 13.3 has 5 default threads (3 purge handles, 1 coordinator, 1 shutdown handler), so thats why we decided to bump up the default number allowed threads for the ``wait_until_load_low`` implementation.

adamchainz:
> MariaDB 13.3 has 5 default threads (3 purge handles, 1 coordinator, 1 shutdown handler), so thats why we decided to bump up the default number allowed threads for the wait_until_load_low implementation.

Put this in the changelog

Also it's 10.3 not 13.3

adamchainz:
Merged in 4913f4df0f23b6addaa918174d9d7bf1bbe0fe74, thanks!

PR #486, opened 2018-06-19
Use temporary database connection during system check

### LINKED ISSUE #485: dangling connection created during system check may become unusable
*Summary*: mysql connection created during system check is not closed timely. This connection might (mysteriously) become unusable afterwards (for reason I could not fully understand for now), which blocks the database access in application logic.

*Description*: I'm using django with celery. Today I noticed that any task that accesses database by Django ORM just hangs forever (i.e. they only start but never finish). After some digging I realized it's database connection that blocks forever and never comes back.

After another really painful debugging process, I figured out that, during django's system checking process, django-mysql created a persistent database connection rather than a temporary one, thus get re-used in application code. But somehow its connection is dropped, which causes any code that implicitly depends on it blocks forever.

Specifically, I think the following code in `django_mysql/checks.py` should use a `BaseDatabaseWrapper.temporary_connection()`, therefore ensure connection is closed cleanly afterwards.

```python
def check_variables(app_configs, **kwargs):
    errors = []

    for alias, connection in mysql_connections():
        with connection.cursor() as cursor: # connection.temporary_connection()
            cursor.execute("""SELECT @@sql_mode,
                                     @@innodb_strict_mode,
                                     @@character_set_connection""")
            # ...
```

(Sorry if this poor bug report seems in a hurry, I'm really tired (it's 4 AM my brain isn't working...) )

*Django Version*: Django 2.0.6

*Database and version used*: mysqld  Ver 5.7.22 for Linux on x86_64 (MySQL Community Server (GPL))

*Version*: Django-MySQL 2.2.2

Naitreey:
Fixes #485.

Summary: Use temporary database connection during system check. This prevents a dangling connection is leaked to application code. In some cases, the dangling connection may become unusable, thus database access in application code gets blocked.

Checklist:

- [x] Docs updated, or N/A

Test Plan: I have tested the changes in my project. Here is how I tested it: I put the following connection checking code in celery task definition body before running any database-related operation:

```python
@shared_task
def task():
    from django.db import connection
    print(connection.connection)
    # ...
```

By starting a fresh celery worker instance and dispatching a task: `task.delay()`, I
can see the following in celery worker log:

```
celery-periodic_1  | [2018-06-19 14:38:59,453: INFO/MainProcess] celery@106f1e5cd0eb ready.
celery-periodic_1  | [2018-06-19 14:39:08,722: INFO/MainProcess] Received task: test.tasks.task[fc05571c-05fd-409d-8b36-489da6e44914]  
celery-periodic_1  | [2018-06-19 14:39:08,726: WARNING/ForkPoolWorker-8] None
celery-periodic_1  | [2018-06-19 14:39:08,757: INFO/ForkPoolWorker-8] Task test.tasks.task[fc05571c-05fd-409d-8b36-489da6e44914] succeeded in 0.03256514499662444s: None
```

The third line of logging indicates no dangling connection was leaked from system check phase.

### DIFF
--- a/django_mysql/checks.py
+++ b/django_mysql/checks.py
@@ -17,7 +17,7 @@ def check_variables(app_configs, **kwargs):
     errors = []
 
     for alias, connection in mysql_connections():
-        with connection.cursor() as cursor:
+        with connection.temporary_connection() as cursor:
             cursor.execute("""SELECT @@sql_mode,
                                      @@innodb_strict_mode,
                                      @@character_set_connection""")


adamchainz:
Awesome, thanks for the detailed PR description! 👌

adamchainz:
Version 2.3.0 out with this fix: https://pypi.org/project/django-mysql/

PR #493, opened 2018-07-25
WIP Test with MySQL 8

### LINKED ISSUE #453: Test on MySQL 8
It's GA now https://mysqlserverteam.com/whats-new-in-mysql-8-0-generally-available/

Normos:
Fixes #453

### DIFF
--- a/tests/testapp/test_locks.py
+++ b/tests/testapp/test_locks.py
@@ -268,7 +268,23 @@ class TableLockTests(TransactionTestCase):
         with TableLock(write=[Alphabet._meta.db_table]):
             assert self.is_locked('default', Alphabet._meta.db_table)
 
-    def test_write_witgit lphabet]):
+    def test_write_with_using(self):
+        Alphabet.objects.using('other').create(a=878787)
+        assert not self.is_locked('other', Alphabet._meta.db_table)
+
+        with TableLock(write=[Alphabet], using='other'):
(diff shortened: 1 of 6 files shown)


tests/testapp/test_locks.py
```diff
@@ -262,6 +262,7 @@ def test_write(self):
         assert Alphabet.objects.count() == 0
 
     def test_write_with_table_name(self):
+        Alphabet.objects.create(a=1)
```
adamchainz: Why is this necessary? You can't lock an empty table on MySQL 8? You could put it in `setUp()` to fix both broken tests and add a comment

tests/testapp/test_models.py
```diff
@@ -119,6 +120,13 @@ def test_approx_count_with_straight_join(self):
 
 class QueryHintTests(TestCase):
 
+    def _skip_for_mysql8(self):
+        if (
+            not connection_is_mariadb(connection) and
+            connection.mysql_version >= (8, 0)
+        ):
+            raise SkipTest("Query cache removed in MySQL 8.0")
```
adamchainz: Nice, but a decorator applied to the tests would be a more conventional way of doing this, imitating `skipUnless`, which we can't use directly as it takes a condition evaluated at import time and we can't connect to DB until runtime

adamchainz:
Replaced by #583, which has just been merged. Thanks for the help, it was a useful start!

PR #530, opened 2019-02-02
Remove support for Django 1.8, 1.9, and 1.10

adamchainz:
* `HISTORY.rst` - add note
* `README.rst` - update list of tested versions
* `setup.py` - specify Django version in `install_requires`, remove classifiers
* Drop all `django.VERSION` based code paths
* Drop `compat` module as it's not doing anything now


yanyixing:
I have a app using django version 1.8, and the lastest django-mysql not support that version. And which django-mysql version support django 1.8 ?

adamchainz:
This PR was released in 2.5.0: https://github.com/adamchainz/django-mysql/blob/master/HISTORY.rst

Therefore, from reading the changelog, versions prior to 2.5.0 support Django 1.8.

I hope you understand you shouldn't be running 1.8 at this point since it is unsupported, the minimum supported version from Django is 1.11 as per https://www.djangoproject.com/download/

Hope that helps,

Adam

PR #539, opened 2019-03-22
Include latest stable Python into the test run

Toshakins:
It's a draft pull request. I open it to check if tests will run on Python 3.7 fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ notifications:
   webhooks: https://coveralls.io/webhook?repo_token=7l8yKtg4su0M31T0eGIaM0tnjvH0Rd5PM
 
 language: python
-python: '3.6'
+python: 
+  - '3.6'
+  - '3.7'
 cache: pip
 
 env:


Toshakins:
Sorry, I thought Travis would test PR by default :( I will do all needed checks offline.

Again sorry for the hassle and opening an incorrect PR.

adamchainz:
I have a plan to test on Python 3.7 myself so no need to do the work yourself

Toshakins:
Thanks for the answer! By the way, I found that one of the projects I involved in successfully uses 3.5.

Again sorry for opening PR without the compliance with CONTRIBUTING guide.

adamchainz:
Yeah technically I aim to support Python 3.4+, I don't have testing there but generally the changes between Python versions are small enough that I'd realize if code was using something not in 3.4. Plus if there was a bug report from an untested Python, I'd fix that. When moving to test with Python 3.7 I aim to revisit this.

PR #545, opened 2019-04-28
Support Django 2.2

adamchainz:
* Add HISTORY note and update installation instructions
* Add 2.2 to tox testing grid
* Update `classifiers` in `setup.py`


adamchainz:
Need to move to Xenial first to get latest SQLite.

adamchainz:
Xenial done in #548

PR #548, opened 2019-04-28
Move Travis testing to Xenial

adamchainz:


adamchainz:
I've just had another look, configuring the MySQL xenial repo is a pain since MySQL 5.6 has disappeared from there. I think using dbdeployer is the answer here, as per https://www.dbdeployer.com/ . Or maybe docker?

PR #596, opened 2019-12-06
Fix GroupConcat to put ORDER BY before SEPARATOR

tuky:
Otherwise you end up with a syntax error if you set both, `separator` and `ordering`!

https://dev.mysql.com/doc/refman/5.7/en/group-by-functions.html#function_group-concat


adamchainz:
Thanks for the fix. I've simplified the test and added a HISTORY note and will merge shortly.

adamchainz:
Released in version 3.3.0, thanks! https://pypi.org/project/django-mysql/3.3.0/

tuky:
Awesome, thanks!

Issue #610, opened 2020-02-09
Faster TransactionTestCase


jonatron:
I had an issue where `TransactionTestCase` was very slow with MySQL. Upon profiling, I found that `TRUNCATE`ing all tables was taking a significant amount of time. I wrote a `TransactionTestCase` SubClass to only `TRUNCATE` tables that aren't empty. Would you like a Pull Request for this?

adamchainz:
Sure, if you’re willing to add docs too!

PR #611, opened 2020-02-10
faster transaction test case

jonatron:
For issue #610 .

### DIFF
--- a/docs/test_cases.rst
+++ b/docs/test_cases.rst
@@ -1,0 +1,14 @@
+.. _test_cases:
+
+==============
+Test Cases
+==============
+
+.. currentmodule:: django_mysql.test.cases
+
+The following can be imported from ``django_mysql.test.cases``.
+
+
+.. class:: FasterTransactionTestCase
+
+    If you have many tables, TransactionTestCase can be slow because Django executes ``TRUNCATE`` on every table after every test. FasterTransactionTestCase will only ``TRUNCATE`` non-empty tables.
--- a/src/django_mysql/test/cases.py
+++ b/src/django_mysql/test/cases.py
@@ -1,0 +1,102 @@
+from __future__ import unicode_literals
+
+import sys
+
+from django.apps import apps
+from django.db import connections, transaction
+from django.db.models.fields import AutoField
+from django.core.management.base import CommandError
+from django.core.management import call_command
+from django.test import TransactionTestCase
+from django.utils import six
+
+
+models_without_auto_inc = []
+
+
+def find_models_without_auto_inc():
+    all_models = apps.get_models()
+    for model in all_models:
+        fields = model._meta.fields
+        has_auto = False
+        for field in fields:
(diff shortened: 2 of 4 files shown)


docs/test_cases.rst
```diff
@@ -0,0 +1,14 @@
...
+==============
+Test Cases
+==============
+
+.. currentmodule:: django_mysql.test.cases
+
+The following can be imported from ``django_mysql.test.cases``.
+
+
+.. class:: FasterTransactionTestCase
```
jonatron: Not that familiar with readthedocs, but class doesn't link to source for me?

src/django_mysql/test/cases.py
```diff
@@ -0,0 +1,102 @@
...
+                    self.serialized_rollback and
+                    hasattr(connections[db_name], '_test_serialized_contents')
+                )
+            )
+            connection = connections[db_name]
+            if "mysql" in connection.settings_dict['ENGINE']:
+                tables = connection.introspection.django_table_names(only_existing=True, include_views=False)
+                schema_name = connection.settings_dict['NAME']
+
+                # auto_increment is reliable, but table_rows isn't
```
jonatron: From what I can tell, this is true, but it'd be good if someone familiar with MySQL internals could confirm.

adamchainz: It's true it's unreliable. Therefore I think this isn't safe in the general case, since a table with a non-auto-inc PK could be reported as empty when not, and then not truncated. You could do `select 1 from x limit 1` on those tables that returned 0. And even `union` together such queries, in batches, to fetch details for multiple tables in one batch, e.g. for two tables:

```
chainz@localhost [14]> (select 't' as has_rows from t limit 1) union (select 't2' as has_rows from t2 limit 1);
+----------+
| has_rows |
+----------+
| t2       |
+----------+
1 row in set (0.001 sec)
```

adamchainz: Also `delete from` can be faster for low row counts, since it doesn't physically recreate the table on disk. With a single row table:

```
chainz@localhost [21]> truncate t2;
Query OK, 0 rows affected (0.005 sec)

chainz@localhost [22]> delete from t2;
Query OK, 0 rows affected (0.000 sec)
```

I think even this optimization would work, using estimated `row_count` - `delete from` for small tables, `truncate` for big ones.

jonatron:
Thanks for taking a look, there's some good further speed improvements. This might be too niche to be merged anywhere, I'll have a think.

adamchainz:
> This might be too niche to be merged anywhere, I'll have a think.

I'd love to see it in Django core!

adamchainz:
I went ahead and made tickets that point back here:

https://code.djangoproject.com/ticket/31275 - "Optimize MariaDB/MySQL sql_flush"

https://code.djangoproject.com/ticket/31276 - "Add 'flush' option to only clear non-empty tables"

PR #614, opened 2020-02-12
Provide Ability to Specify Index Hints for Update Queries

fredpalmer:
# Summary

The MySQL grammar for `UPDATE` and `SELECT` share the same token specification for `table_reference`, i.e.:

```
tbl_name [[AS] alias] [index_hint]
```

This enhancement also rewrites queries to include index hints for `UPDATE` statements.


adamchainz:
Great! Please also test at the ORM level and add docs!

adamchainz:
Clsoing due  to inaction. If you come back to this please also:
* revert the unnecessary removal of verbose regex mode
* add a history note
* add documentation with an example

PR #621, opened 2020-03-03
Prevent JSONField from adding CAST(... AS JSON) for str/int/float

### LINKED ISSUE #480: DB index is not used on some queries
### Summary: 

django-mysql allows to perform a query on JSON encapsulated fields. To prevent performance issues, we created an index as explained [here](https://mysqlserverteam.com/indexing-json-documents-via-virtual-columns/).
So we expect the index to be used when running this query:
```
TableDoc.objects.filter(field__encaspulated_field="value")
```
But it does not, as you can see below (the sql query is the one generated by django):
```
EXPLAIN SELECT * FROM `tabledoc` 
WHERE JSON_EXTRACT(`tabledoc`.`field`, '$.encaspulated_field') = (CAST('\"value\"' AS JSON));
```
id|	select_type|	table|	partitions|	type|possible_keys|	key|	key_len|	ref|	rows|	filtered|	Extra
----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- |  ----- | ----- | ----- | ----- |  
1|SIMPLE	|tabledoc	|NULL	|ALL|	encaspulated_field|	NULL|	NULL|	NULL|	71763|	10.00|	Using where; Using filesort

**When modifying the query to use naively the value without JSON CAST, the index is well used.**
```
EXPLAIN SELECT * FROM `tabledoc` 
WHERE JSON_EXTRACT(`tabledoc`.`field`, '$.encaspulated_field') = "value");
```
id|	select_type|	table|	partitions|	type|possible_keys|	key|	key_len|	ref|	rows|	filtered|	Extra
----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- |  ----- | ----- | ----- | ----- |  
1|SIMPLE	|tabledoc	|NULL	|ALL|	encaspulated_field|	NULL|	NULL|	NULL|	1|	10.00|	Using where; Using filesort

I'm not sure if it is a bug and if it has been solved on more recent django/django-mysql versions but I would appreciate your feedbacks.

Thanks :)

### Configuration

Django Version: *Django 1.9.1*

Database and version used: *MySQL 5.7*

Version: *Django-MySQL 1.1.0*

arnau126:
Fixes #480 .
Only cast to json the right hand side expression if it is a dict, a list or a tuple.

Example:
Before:
`WHERE JSON_EXTRACT(tabledoc.field, '$.sub_field') = (CAST('\"value\"' AS JSON));`
After:
`WHERE JSON_EXTRACT(tabledoc.field, '$.sub_field') = ("value");`

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,8 @@ Pending
 
 * Prevent ``collections.abc.Sequence`` warning.
 * Drop Django 1.11 support. Only Django 2.0+ is supported now.
+* Prevent ``JSONField`` from adding ``CAST(... AS JSON)`` to queries against
+  ``str``, ``int``, and ``float`` types.
 
 3.3.0 (2019-12-10)
 ------------------
--- a/src/django_mysql/models/lookups.py
+++ b/src/django_mysql/models/lookups.py
@@ -2,7 +2,7 @@ import json
 from collections.abc import Sequence
 
 import django
-from django.db.models import CharField, Lookup, Transform
+from django.db.models import CharField, Lookup, Transform, Value
 from django.db.models.lookups import (
     BuiltinLookup,
     Exact,
@@ -53,7 +53,10 @@ class JSONLookupMixin:
         def get_prep_lookup(self):
             value = self.rhs
             if not hasattr(value, "resolve_expression") and value is not None:
-                return JSONValue(value)
+                if isinstance(value, (str, int, float)) and not isinstance(value, bool):
+                    return Value(value)
+                else:
+                    return JSONValue(value)
             return super().get_prep_lookup()
 
     else:
@@ -61,7 +64,10 @@ class JSONLookupMixin:
         def get_prep_lookup(self):
             value = self.rhs
             if not hasattr(value, "_prepare") and value is not None:
-                return JSONValue(value)
+                if isinstance(value, (str, int, float)) and not isinstance(value, bool):
+                    return Value(value)
+                else:
+                    return JSONValue(value)
             return super().get_prep_lookup()
 
 
--- a/tests/testapp/models.py
+++ b/tests/testapp/models.py
@@ -280,6 +280,8 @@ class JSONModel(Model):
     ) and connection._nodb_connection.mysql_version >= (5, 7):
         attrs = JSONField(null=True)
 
+    name = CharField(max_length=3)
+
     def __unicode__(self):
         return str(json.dumps(self.attrs))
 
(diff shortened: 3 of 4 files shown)


adamchainz:
I think it would be safer to invert the logic and use `Value` only for some white-listed cases such as `int`, `Col`, etc. ?

arnau126:
Like this?
```python
    if isinstance(value, (int, float, str)) and not isinstance(value, bool):
        return Value((value)
    else:
        return JSONValuevalue)
```

Notice that `bool` must be casted, but as it is a subclass of `int` we need to add this second condition.

I don't know which type is `Col`.

arnau126:
@adamchainz , any thoughts?

arnau126:
@adamchainz , what can I do to make this PR go ahead?

adamchainz:
Merged and released in 3.4.0. 🎉

adamchainz:
Sorry for the delay. Next time please include a changelog note.

alexgmin:
I'm very late for this, but after some investigation I found out that this pull request was a breaking change, at least in Mysql 5.7

I have a column with these types of values:
```json
{
    "foo": "bar",
    "list": [
        "aaa",
        "bbb"
    ]
}
```
Before, with automatic Json casting, this worked:
`Model.objects.filter(column__list__contains='aaa')`
After this I have to cast the value to a string myself:
`Model.objects.filter(column__list__contains='"aaa"')`

Note that for this query to fail there has to be a row with the list inside, otherwise Mysql optimizes it away.

arnau126:
Yes, you're right.

The problem is that the `=` operator does not work in the same way than JSON_CONTAINS.

Example: 
```mysql
SET @doc1 = '{"a": "x", "b": 2, "c": {"d": 3}}';
SELECT JSON_EXTRACT(@doc1, '$.a') = cast('"x" as json');  # before this PR
SELECT JSON_EXTRACT(@doc1, '$.a') = 'x';  # after this PR
```
In both cases the SELECT returns `1`, and that's why we changed it.

But what I didn't know is that JSON_CONTAINS works in a different way with scalar values:
```mysql
SET @doc1 = '{"a": "x", "b": 2, "c": {"d": 3}}';
SELECT JSON_CONTAINS(@doc1, cast('"x"' as json), '$.a');
SELECT JSON_CONTAINS(@doc1, 'x', '$.a');
```
The first one returns `1` as expected, but the second one raises an error:
> Error Code: 3141. Invalid JSON text in argument 2 to function json_contains: "Invalid value." at position 0.


So, although `=` operator and JSON_CONTAINS work in a different way for scalar values, both classes (JSONExact and JSONContains) inherit from the same superclass: JSONLookupMixin.

I think the solution is make JSONContains inherit from another superclass. A superclass whose `get_prep_lookup` function always dumps and casts values to json.

arnau126:
Fixed in #668 .

alexgmin:
Thank you for the quick fix!

PR #622, opened 2020-03-03
Implement JSONIn lookup.

arnau126:
In both [mysql5.7](https://dev.mysql.com/doc/refman/5.7/en/json.html) and [mysql8.0](https://dev.mysql.com/doc/refman/8.0/en/json.html) documentation says:

> The following comparison operators and functions are not yet supported with JSON values: BETWEEN, **IN()**, GREATEST(), LEAST(). A **workaround** for the comparison operators and functions just listed is to cast JSON values to a native MySQL numeric or string data type so they have a consistent non-JSON scalar type.

In this PR I've implemented this workaround for the **In** operator.

Note that the workaround is achieved by doing `json.dumps` to the values in the right-hand-side. This is not possible if the values are expressions or subqueries.

### DIFF
--- a/src/django_mysql/models/lookups.py
+++ b/src/django_mysql/models/lookups.py
@@ -8,6 +8,7 @@ from django.db.models.lookups import (
     Exact,
     GreaterThan,
     GreaterThanOrEqual,
+    In,
     LessThan,
     LessThanOrEqual,
 )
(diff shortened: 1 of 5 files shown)


src/django_mysql/models/lookups.py
```diff
@@ -69,6 +70,25 @@ class JSONExact(JSONLookupMixin, Exact):
...
+            return super(JSONIn, self).get_prep_lookup()
+        prepared_values = []
+        for rhs_value in self.rhs:
+            if hasattr(rhs_value, "resolve_expression"):
+                pass
+            else:
+                if isinstance(rhs_value, (dict, list, tuple)):
+                    rhs_value = JSONValue(rhs_value)
+                else:
+                    rhs_value = Value(json.dumps(rhs_value, allow_nan=False))
```
adamchainz: same here as I comented on #621 - might be better to whitelist types that can go through `Value`. Also it would be better to use a `Value` subclass rather than `json.dumps` directly here - perhaps rename the original `CastJSONValue` and this one `JSONValue`?

arnau126: renaming and changelog done, @adamchainz

arnau126: and PR description edited

arnau126:
Rebased.
Tested with mysql8.
Eventually, we don't need to rename the current JSONValue to CastJSONValue and implement a new one that does not cast values, because in mysql8 all values need to be casted.

adamchainz:
Hi @arnau126 

I'm sorry I ignored this PR so far. I haven't been in the right headspace to dive into the specifics of MySQL JSON and the 'IN' operator.

I've pushed documentation and a test for chained lookups.

FYI I'm going to deprecate the Django-MySQL JSONField with the upcoming release of Django 3.1 that has a universal JSONField, including MariaDB support which is not something I ever got working here. It doesn't support `__in` at the moment - [see docs](https://docs.djangoproject.com/en/3.1/topics/db/queries/#querying-jsonfield). If you want to contrbiute a ticket there, it'd be welcome.

If you've not contributed to Django before, here's the start page: https://docs.djangoproject.com/en/dev/internals/contributing/

Thanks,

Adam

arnau126:
Thanks for the documentation.
I already have some tests for chained lookups in `ExtraLookupsQueryTests.test_in_lookup()`. Thank you anyway.

Thanks for the info about Django3.1 and the Django-MySQL JSONField deprecation. I will take a look at contributing in Django.

PR #624, opened 2020-03-06
docs: Fix simple typo, overriden -> overridden

timgates42:
There is a small typo in docs/model_fields/json_field.rst.

Should read `overridden` rather than `overriden`.


adamchainz:
I guess when I wrote this I was in Middle English mode: https://en.wiktionary.org/wiki/overriden

adamchainz:
Thanks!

PR #642, opened 2020-04-18
Use volatile MariaDB/MySQL configuration

adamchainz:
As per http://www.tocker.ca/2013/11/04/reducing-mysql-durability-for-testing.html


adamchainz:
Not worth the effort.

PR #646, opened 2020-04-26
Feature: return iterables from groupconcat

javiplx:


adamchainz:
Thank you, but please add tests! How are you verifying this already? Also rebase on the latest master, you have conflicts.

javiplx:
This was in fact for our own use, with an older django version

PR #647, opened 2020-04-26
Select proper default for `output_field` based on distinct value

javiplx:
Will return a set when distinct is enabled, and a list otherwise

### DIFF
--- a/src/django_mysql/models/aggregates.py
+++ b/src/django_mysql/models/aggregates.py
@@ -1,4 +1,5 @@
 from django.db.models import Aggregate, CharField
+from django_mysql.models.fields import SetCharField, ListCharField
 
 
 class BitAnd(Aggregate):
@@ -24,8 +25,10 @@ class GroupConcat(Aggregate):
     ):
 
         if "output_field" not in extra:
-            # This can/will be improved to SetTextField or ListTextField
-            extra["output_field"] = CharField()
+            if distinct:
+                extra['output_field'] = SetCharField(CharField())
+            else:
+                extra['output_field'] = ListCharField(CharField())
 
         super().__init__(expression, **extra)
 


tests/testapp/test_aggregates.py
```diff
@@ -104,14 +103,14 @@ def test_separator_big(self):
...
-        concatted_ids = ",".join([str(self.jk.id + 1), str(self.grisham.id + 1)])
+        concatted_ids = [str(self.jk.id + 1), str(self.grisham.id + 1)]
         assert out == {"tids": concatted_ids}
 
     def test_application_order(self):
         out = Author.objects.exclude(id=self.shakes.id).aggregate(
             tids=GroupConcat("tutor_id", distinct=True)
         )
-        assert out == {"tids": str(self.shakes.id)}
+        assert out == {"tids": set(map(str,[self.shakes.id]))}
```
adamchainz, line 113:
```suggestion
        assert out == {"tids": {str(self.shakes.id)}}
```

javiplx: Thanks. I'm quite old school and sets are somewhat new to me ...

javiplx:
I believe that all functional tests succeed, but have no idea neither why the py38-multilint complains nor how to fix that particular error.

adamchainz:
The error:

```
py38-codestyle run-test: commands[0] | multilint
Running black check
would reformat /home/runner/work/django-mysql/django-mysql/src/django_mysql/models/aggregates.py
Oh no! 💥 💔 💥
1 file would be reformatted, 90 files would be left unchanged.
black failed
```

Run [black](http://black.readthedocs.io/) on the file to fix that.

adamchainz:
So thinking about how we release this - it's a small breaking change. I think it will need a major version bump to communicate this, so I'd like to batch it with some other breaking changes I've not got around to making.

adamchainz:
@javiplx It'll be faster if you run `isort` and `black` locally, rather than iterating against them on CI :) If you don't have python 3.8 locally, you should be able to run them with 3.6 or 3.7 just fine.

javiplx:
Finally !!!!  
Regarding releasing, I agree with the mayor approach because this is not backwards compatible. In any case we'll make an internal release based on 3.2.0, because we're still running django 1.11

PR #651, opened 2020-05-03
Tested on MySQL v8.0.17

sudarshandk:
Running the ./manage.py check results in the below output:

System check identified no issues (0 silenced)


adamchainz:
I'm afraid that doesn't run the tests at all. We can't just document it as working because the (few) system checks pass.

#583 is my start of running the tests on MySQL 8 and needs completing.

PR #656, opened 2020-05-19
Added missing JSONRemove util function

Ashish-Bansal:
### Description

Self exp.

Let me know if something like this is acceptable. I'll add docs and tests for the same then. 

Ref - https://dev.mysql.com/doc/refman/5.7/en/json-modification-functions.html#function_json-remove

### DIFF
--- a/src/django_mysql/models/functions.py
+++ b/src/django_mysql/models/functions.py
@@ -285,13 +285,16 @@ class JSONValue(Func):
 class JSONRemove(Func):
     function = 'JSON_REMOVE'
 
-    def __init__(self, expression, path):
+    def __init__(self, expression, paths):
         from django_mysql.models.fields import JSONField
 
-        if not hasattr(path, 'resolve_expression'):
-            path = Value(path)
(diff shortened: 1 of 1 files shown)


adamchainz:
Hi

Django 3.1 includes a universal JSONField. It's likely I'll be deprecating this one in due course, but I'm happy to merge and release such a function if it helps you now.

However the PR needs a lot more work. Please add:

* Tests - until coverage is 100%
* Documentation
* A release note

See previous PR's / commits and existing code for examples. It's also worth reading the django contribution guide for general guidance on how best to contribute to open source projects.

Thanks,

Adam

Ashish-Bansal:
> Django 3.1 includes a universal JSONField. It's likely I'll be deprecating this one in due course

[0] - I see. If that's the case, I don't think it's worth adding. In the meanwhile, I'll add it directly in our project. 

Closing this PR. Thanks!

[0] https://docs.djangoproject.com/en/dev/releases/3.1/#jsonfield-for-all-supported-database-backends

Issue #643, opened 2020-04-25
rewrite_query API stability/programmatic setting


jheld:
Summary: *Consider making rewrite setting/flow programmatic*

`DJANGO_MYSQL_REWRITE_QUERIES` setting is static. We do understand why that's the case historically (powerful but easy to turn off). After reading through the code it seems that the initial small overhead to each query is a string comparison to check for the rewrite marker in the sql, when the setting is on. Although the string check generally should be quite fast (sql queries in mysql in general cannot go above 1GB anyway, and most likely are much smaller than that), the overhead does indeed still pose a performance risk for larger queries (and it's not always known if/when they could occur).

We are considering attempting to add support to make the setting allowed to be a callable. So for instance, if we were to make a django request middleware (considering certain user permissions, or certain APIs), we could make the setting on/off during the execution of that particular request.

Now, having read the source code for rewrite_query and the monkey_patches, I'm not certain that a middleware would be effective (perhaps you can verify my question).

On the other hand, middleware or not, if we could guarantee a sort of middle level public API for what the monkey patching does (checks the setting but also alters the sql if it matches the rewrite logic, based on wrapping the cursor's execute function), then that potentially could allow a ton of official flexibility for projects that wish to have programmatic access to rewrite, if they so choose, and without risking a breakage in compatibility with this library should something change (since it is only a somewhat exposed function, incidentally public, rather than explicitly).

There may be methods to hack around this (perceived limitation), but I am hoping that we can find a more official path that embraces flexibility, rather than being unaware that projects might be relying on implied behavior.

What are your thoughts so far?

adamchainz:
> the overhead does indeed still pose a performance risk for larger queries (and it's not always known if/when they could occur).

First, I'll address this. I don't believe there's much overhead, and you haven't provided any evidence that it's noticeable. If you do any profiling I think you'll find the overhead imperceptible compared to many other parts of the Django ORM.

Searching for the marker in a 1000 character string is 200 nanoseconds:

```
In [1]: string = 1_000 * "B"

In [2]: marker ="/*QueryRewrite':"

In [3]: %timeit marker in string
207 ns ± 0.796 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Many queries are even shorter so the overhead will be less. Yes, there's a settings lookup and a function call missing from this profile, but that should be minimal.

[...]

PR #658, opened 2020-05-20
Allow callable for rewrite setting

tredzko:
This was touched on in #643. The response to that was looking into middleware for performance/flow reasons. This MR addresses the programmatic enabling/disabling of rewriting queries so that a callable can be provided instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,7 @@ output/*/index.html
 
 # Sphinx
 docs/_build
+
+# Mac
+.DS_Store
+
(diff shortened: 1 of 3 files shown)


adamchainz:
This PR doesn’t seem like it would do hat you want. The callable is only evaluated once, in patch, at application start. You may as well evaluate any such function call in settings instead.

And anyway, running a callable per query is not something I want to add. Checking for and evaluating the callable is likely to be much slower than the “string in string” check. And if you really want to use a function call per query, Python already allows you to: you can set the setting to an object with a custom __bool__ that does whatever you want.

Also, you haven’t responded to my profiling with any profiling data of your own. Please provide profiles of a real world scenario to show that this wrapper has an actual performance impact. Otherwise it’s just guesswork.

PR #660, opened 2020-05-24
Test with Django 3.1

adamchainz:


adamchainz:
Pending discussion about the `.delete()` change in [django ticket 31624](https://code.djangoproject.com/ticket/31624). Without a fix for that, `TableLock` is much less useful.

adamchainz:
Awaiting next alpha release.

PR #665, opened 2020-05-31
Move query rewriting to database instrumentation

### LINKED ISSUE #644: Move query rewriting to database instrumentation
Summary: Rather than monkeypatching `CursorWrapper.execute`, use [database instrumentation](https://docs.djangoproject.com/en/3.0/topics/db/instrumentation/) to install the hook.

The setting can then always install it with the same technique as https://github.com/scoutapp/scout_apm_python/blob/639b7ac6a6a859e17db5c2e6b4d6ce2366ddd7a4/src/scout_apm/django/instruments/sql.py#L71 . And users can optionally install it only for certain code paths by installing it using `with connection.execute_wrapper(rewrite_queries):` (would need documenting).

adamchainz:
Fixes #644.


jheld:
Thank you for implementing this!

PR #667, opened 2020-06-03
Add more tests for the JSONContains lookup.

arnau126:
I've added some more test to ensure that #621 has not broken JSONContains lookup.


arnau126:
> I've added some more test to ensure that #621 has not broken JSONContains lookup.

#621 did break  JSONContains lookup...

PR #668, opened 2020-06-03
Fix bug in JSONContains lookup related with scalar values.

arnau126:
#621 broke the "contains" lookup, as the mysql fuction "JSON_CONTAINS" needs all values to be casted to json, even scalar values.

Example (mysql 5.7)
```mysql
SET @doc1 = '{"a": "x", "b": 2, "c": {"d": 3}}';
SELECT JSON_CONTAINS(@doc1, cast('"x"' as json), '$.a');  #  returns 1
SELECT JSON_CONTAINS(@doc1, 'x', '$.a');  # ERROR
```
The error raised by the second SELECT is:
> Error Code: 3141. Invalid JSON text in argument 2 to function json_contains: "Invalid value." at position 0.

This behavior differs from the one from the `=` mysql operator, which works with both versions:
```mysql
SET @doc1 = '{"a": "x", "b": 2, "c": {"d": 3}}';
SELECT JSON_EXTRACT(@doc1, '$.a') = cast('"x" as json');  # returns 1
SELECT JSON_EXTRACT(@doc1, '$.a') = 'x';  # returns 1
```

### DIFF
--- a/src/django_mysql/models/lookups.py
+++ b/src/django_mysql/models/lookups.py
@@ -101,8 +101,28 @@ class JSONContainedBy(Lookup):
         return "JSON_CONTAINS({}, {})".format(rhs, lhs), params
 
 
-class JSONContains(JSONLookupMixin, Lookup):
+class JSONContains(Lookup):
     lookup_name = "contains"
+
+    if django.VERSION >= (3, 0):
+
+        def get_prep_lookup(self):
+            value = self.rhs
+            if hasattr(value, "resolve_expression"):
+                raise ValueError(
+                    "JSONField's 'contains' lookup only works with literal values"
+                )
+            return JSONValue(value)
+
+    else:
+
+        def get_prep_lookup(self):
+            value = self.rhs
+            if hasattr(value, "_prepare"):
+                raise ValueError(
+                    "JSONField's 'contains' lookup only works with literal values"
+                )
+            return JSONValue(value)
 
     def as_sql(self, qn, connection):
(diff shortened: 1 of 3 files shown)


src/django_mysql/models/lookups.py
```diff
@@ -101,9 +101,29 @@ def as_sql(self, qn, connection):
...
+
+    else:
+
+        def get_prep_lookup(self):
+            value = self.rhs
+            if hasattr(value, "_prepare"):
+                raise ValueError(
+                    "JSONField's 'contains' lookup only works with literal values"
+                )
+            return JSONValue(value)
```
adamchainz, line 125: Why block expressions, rather than casting their results? This is backwards incompatible.

arnau126: Expressions never worked with the contains lookup.

`JSONModel.objects.filter(attrs__a__contains=F("name"))`  was translated into
```mysql
SELECT ...
FROM `testapp_jsonmodel`
WHERE JSON_CONTAINS(JSON_EXTRACT(`testapp_jsonmodel`.`attrs`, $.a), (`testapp_jsonmodel`.`name`))
```
and raised:
`django.db.utils.OperationalError: (3141, 'Invalid JSON text in argument 2 to function json_contains: "Invalid value." at position 0.')`

Maybe there's a way to make it work, but this PR is not backwards incompatible.

adamchainz: Okay cool, we're just raising a more sensible error.

Issue #671, opened 2020-06-10
Support  ON UPDATE CURRENT_TIMESTAMP ( what I'll to make improvement) #674


KimSoungRyoul:
Hi 
I want to use these Feature with Django
https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html

---

### To be
~~~Python
class Abc(models.Model):
    name = models.CharField(max_length=32)
    modified_at = models.DateTimeField(auto_now=True)

~~~
if i execuete below Queryset 
~~~Python
Abc.object.filter(id=1).update(name="qwer")
~~~
`modified_at`Field Does Not be updated 

---

### As Is

Mysql Support these Feature with DDL 
https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html

##### if you don't mind  , can I suggest  create DateTimeField with `on_update_current_timestamp` option   ?
~~~Python
from django_mysql import models as mysql_models

class Abc(models.Model):
    name = models.CharField(max_length=32)
    modified_at = mysql_models.DateTimeField(
                          auto_now_add=True, 
                          on_update_current_timestamp=True # new param "on_update_current_timestamp"
                  ) 

~~~

when `on_update_current_timestamp` is True, 
DDL will be generated like a
~~~SQL
create table pratice_111_db.hello11_asdf
(
    name varchar(32) not null,
    a_datetime datetime(6) not null ON UPDATE CURRENT_TIMESTAMP, # `on_update_current_timestamp=True
    id         int auto_increment
        primary key
);
~~~

---

adamchainz:
I don't have time to write something here but you should be able to customize `DateTimeField` to do that. I'd be interested if you had an implementation.

KimSoungRyoul:
https://github.com/adamchainz/django-mysql/pull/674

PR #674, opened 2020-06-20
Create datetimefield for on update option #671 

KimSoungRyoul:


adamchainz:
Why did you close this?

KimSoungRyoul:
I thought document wasn't  enough 
I need more addition and reinforce testcase

src/django_mysql/models/fields/datetime.py
```diff
@@ -0,0 +1,78 @@
+from django.core import checks
+from django.db.models import DateTimeField as DjangoDateTimeField
+from django.db.models.fields import DateTimeCheckMixin as DjangoDateTimeCheckMixin
+from django.utils import timezone
+
+
+class DateTimeCheckMixin(DjangoDateTimeCheckMixin):
```
adamchainz, line 7: There's no need to subclass the mixin to replace this method in the new `DateTimeField` class.

src/django_mysql/models/fields/datetime.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, on_update_current_timestamp=False, **kwargs):
+        self.on_update_current_timestamp = on_update_current_timestamp
+        super(DateTimeField, self).__init__(**kwargs)
+
+    def db_type_suffix(self, connection):
+        db_type_suffix = super().db_type_suffix(connection)
+        if db_type_suffix is None:
+            db_type_suffix = "ON UPDATE CURRENT_TIMESTAMP(6)"
+        else:
+            db_type_suffix += "ON UPDATE CURRENT_TIMESTAMP(6)"
```
adamchainz, line 56: Wouldn't you want a space before "ON" ?

src/django_mysql/models/fields/datetime.py
```diff
@@ -0,0 +1,78 @@
...
+        return db_type_suffix
+
+    def pre_save(self, model_instance, add):
+        if self.auto_now or self.on_update_current_timestamp:
+            value = timezone.now()
+            setattr(model_instance, self.attname, value)
```
adamchainz, line 62: it's a bit disingenous to set the attribute to the current time in python when it's updated in the database, because the two clocks may be out of sync.

tests/testapp/test_datetimefield.py
```diff
@@ -0,0 +1,125 @@
...
+        assert s1.datetime2 == s2.datetime2
+        assert s1.datetime3 == s2.datetime3
+
+
+class TestCheck(TestCase):
+
+    if django.VERSION >= (2, 2):
+        databases = ["default", "other"]
+    else:
+        multi_db = True
```
adamchainz, line 85: We are only testing django 2.2.+ since #673 so please only use `databases = `

src/django_mysql/models/fields/datetime.py
KimSoungRyoul: i confirmed ```django.db.backend.mysql.DatabaseSchemaEditor & DatabaseWrapper ```
DDL is  Generated with end space 

but  i agree to  a space before "ON"  is more safety

src/django_mysql/models/fields/datetime.py
KimSoungRyoul: I intended .save() depend on  **current time in python**
and QuerySet.update() depend on  **MYSQL current_timestamp** due to ``` on_update_current_timestamp=True ```


#### because
we can not control ```field.defualt``` Option in Django Field  & Update SQL 
it's controlled in ``` django.db.backend.*.schemas.XXXDatabaseSchemaEditor```
and
Update SQL  should ignore `DateTImeField(on_update_current_timestamp=True)`
When UPDATE SQL execution

we need to  override `django MYSQLDatabaseSchemaEditor`
https://github.com/django/django/blob/master/django/db/backends/mysql/schema.py
And django.QuerySet.update()

but I thought  it's too huge change Just to modify one little thing(DateTimeField)

I agree `it's a bit disingenous` 
but  I try to make coherent action
` .save()    depend on  current time in python`
`.update() depend on  MYSQL current_timestamp`

With my short insight
If  develop DateTimeField with Perfectly depended MYSQL current_timestamp
We need to refactor Django Backend schemas.DatabaseSchemaEditor
 that  DDL is depended

adamchainz: I don't see what DDL needs modifying for this?

src/django_mysql/models/fields/datetime.py
adamchainz: But db_type_suffix is always `None` there already

src/django_mysql/models/fields/datetime.py
KimSoungRyoul: I'll think about it more.

KimSoungRyoul:
####  In my short opinion, I realized that there is no solution to get current time with only INSERT & Update
#### django connection could not know database's real time(current timp)
#### so even if solution is existed , It's not django-mysql's territory.
#### Rather, it is related to mysql-client feature


#### thank you for your review 

#### i will close this PullRequest

PR #700, opened 2020-07-24
Fix RemovedInDjango40Warning messages for django.utils.translation

jheld:
small adjustment to remove the warning since all supported versions have this as the favored usage.


adamchainz:
Released in 3.8.0, thanks.

PR #714, opened 2020-10-07
Fix form JSONField for to not use the ensure_ascii flag

ZhaoQi99:
```python
>>> import json
>>> print json.dumps('中国')
"\u4e2d\u56fd"
```
`json.dumps` use ASCII encoding by default when serializing Chinese.
So when we edit a `JsonField` which contains Chinese character in Django admin,it will appear in ASCII characters.
Adding `ensure_ascii=False` param will fix this problem.

### DIFF
--- a/src/django_mysql/forms.py
+++ b/src/django_mysql/forms.py
@@ -283,4 +283,4 @@ class JSONField(forms.CharField):
     def prepare_value(self, value):
         if isinstance(value, InvalidJSONInput):
             return value
-        return json.dumps(value)
+        return json.dumps(value, ensure_ascii=False)
--- a/tests/testapp/test_forms.py
+++ b/tests/testapp/test_forms.py
@@ -269,11 +269,14 @@ class TestJSONField(SimpleTestCase):
         field = JSONField()
         assert field.prepare_value({"a": "b"}) == '{"a": "b"}'
         assert field.prepare_value(["a", "b"]) == '["a", "b"]'
+        assert field.prepare_value(["你好，世界", "😀🐱"]) == '["你好，世界", "😀🐱"]'
         assert field.prepare_value(True) == "true"
         assert field.prepare_value(False) == "false"
         assert field.prepare_value(3.14) == "3.14"
         assert field.prepare_value(None) == "null"
         assert field.prepare_value("foo") == '"foo"'
+        assert field.prepare_value("你好，世界") == '"你好，世界"'
+        assert field.prepare_value("😀🐱") == '"😀🐱"'
 
     def test_redisplay_wrong_input(self):
         """


adamchainz:
Thanks. Does the model field need changing too?

Btw Django 3.1 has a JSONField - [model](https://docs.djangoproject.com/en/3.1/ref/models/fields/#jsonfield) / [form](https://docs.djangoproject.com/en/3.1/ref/forms/fields/#jsonfield) . I'd recommend using that if you can. Can you check the bug does not exist there? (It shouldn't - I think it is better tested than this one.)

ZhaoQi99:
> Thanks. Does the model field need changing too?
> 
> Btw Django 3.1 has a JSONField - [model](https://docs.djangoproject.com/en/3.1/ref/models/fields/#jsonfield) / [form](https://docs.djangoproject.com/en/3.1/ref/forms/fields/#jsonfield) . I'd recommend using that if you can. Can you check the bug does not exist there? (It shouldn't - I think it is better tested than this one.)

I just tested Django's JSONField in `Django==3.1.2`,and this problem still exists.
After I add `ensure_ascii=False` in [django/forms/fields.py#L1261](https://github.com/django/django/blob/3d4ffd1ff0eb9343ee41de77caf6ae427b6e873c/django/forms/fields.py#L1261),it can display normaly.

image
image

ZhaoQi99:
Then I read the source code of django tests - [tests/forms_tests/field_tests/test_jsonfield.py](https://github.com/django/django/blob/3d4ffd1ff0eb9343ee41de77caf6ae427b6e873c/tests/forms_tests/field_tests/test_jsonfield.py#L29) , and It seems that Chinese is not considered.

xiekeyi98:
I wonder if this is because internationalization was not considered? If we change this, does it cause any other effect in the MySQL section? For example, if we insert this JSON case of yours into an ASCII table in MySQL?

BTW, would you mind adding some emoji, Japanese, etc.?

ZhaoQi99:
> I wonder if this is because internationalization was not considered? If we change this, does it cause any other effect in the MySQL section? For example, if we insert this JSON case of yours into an ASCII table in MySQL?
> 
> BTW, would you mind adding some emoji, Japanese, etc.?

No,the code which I chaged only influences the display in Django admin's form, and it will not cause any effect about MySQL.

ZhaoQi99:
I have reported this bug in Django's Ticket system. @adamchainz 
https://code.djangoproject.com/ticket/32080

adamchainz:
Thanks for this, and the fix over in Django core! ( https://github.com/django/django/pull/13512 )

adamchainz:
Released in version 3.9.0 ! 🥳

PR #734, opened 2020-12-09
Add some type hints

adamchainz:
Playing around with monkeytype.

Used this `monkeytype_config.py`:

```
import os
from contextlib import contextmanager

from monkeytype.config import DefaultConfig

class MyConfig(DefaultConfig):
    @contextmanager
    def cli_context(self, command):
        os.environ["DJANGO_SETTINGS_MODULE"] = "tests.settings"

        import django

        django.setup()
        yield

CONFIG = MyConfig()
```

And this modification to `tox.ini` to run the tests with tracing:

```
setenv = MONKEYTYPE_TRACE_MODULES=django_mysql
commands = monkeytype run -m pytest {posargs}
```

Only captured ~50% of calls. Needed a bunch of manual tweaking.

### DIFF
--- a/src/django_mysql/apps.py
+++ b/src/django_mysql/apps.py
@@ -1,5 +1,8 @@
+from typing import Any, Callable, Iterable
+
 from django.apps import AppConfig
 from django.conf import settings
+from django.db.backends.base.base import BaseDatabaseWrapper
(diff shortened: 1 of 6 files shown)


adamchainz:
Gonna add type hints manually.

PR #737, opened 2020-12-09
Replace string concatenation method

### LINKED ISSUE #598: Try += for string concatenation
Summary: 

Paul Ganssle's recent article showed that using `+=` rather than `''.join` can be faster for string concatenation in Python. https://blog.ganssle.io/articles/2019/11/string-concat.html

For lots of the SQL generation in Django-MySQL I suspect this also holds and could be a nice little improvement.

opeerator:
Fixes #598 

Summary:

Using `+=` rather than `.join` can be faster for string concatenation in Python, according to [this article](https://blog.ganssle.io/articles/2019/11/string-concat.html).

Solution:

replacing all of the `.join` methods in the src folder to `+=` for string concatenation.

### DIFF
--- a/src/django_mysql/cache.py
+++ b/src/django_mysql/cache.py
@@ -277,11 +277,11 @@ class MySQLCache(BaseDatabaseCache):
             value, value_type = self.encode(value)
             params.extend((made_key, value, value_type, exp))
 
-        val_c = ""            
+        val_c = ""
         for _ in data:
             if val_c:
                 val_c += ","
-            val_c += "(%s, %s, %s, %s)"            
+            val_c += "(%s, %s, %s, %s)"
 
         query = self._set_many_query.replace(
             "{{VALUES_CLAUSE}}", val_c).format(table=table)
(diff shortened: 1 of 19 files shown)


adamchainz:
This issue was a bit of a "note to self" and a suspicion, not a fact. Did you time any of the instances you changed?

opeerator:
Oh, I see. I thought we might have a similar behavior like the article mentions here as well. I'm trying to time some of the instances, but it does not always favor concat or join concretely.

adamchainz:
I'd like to see concrete numbers from `%timeit` on ipython before merging any change.

opeerator:
Here you go, yeah it doesn't seem much of an improvement in every case. In cases that we have `.format()`, `.join` seems a better option. :/

[time_modules.pdf](https://github.com/adamchainz/django-mysql/files/5680064/time_modules.pdf)

adamchainz:
Thanks for this. Yeah I don't think any of the cases are significant enough to warrant a change here.

It's likely there are similar paths in Django that can be optimized the same way, and they'd have higher impact since it's much more popular. Perhaps you could go hunting there? We're more interested in saving microseconds there since they multiply out to a bigger scale.

Also check out [djangobench](https://github.com/django/djangobench) which is Django's performance benchmarking suite, currently undergoing some improvement work.

PR #761, opened 2021-03-13
Support for non-ascii key.

muzige2000:


adamchainz:
Hi

This code is failing tests

Also, I'm no longer maintaing the JSON field implementation here - please upgrade to Django 3.1 and use the implementation it contains.

Thanks,

Adam

PR #786, opened 2021-06-09
Fix index hints for tables with aliases

hjaarnio:
If django added an alias to a table in subquery or join the SQL produced was invalid.
eg.
```
WHERE (EXISTS (
    SELECT (1) FROM `table_name` U0
    WHERE U0.`outer_id` = `outer_table`.`id`
))
```
the USE INDEX hint would be added between `table_name` and U0, which is
invalid SQL.
Change to put USE_INDEX after the U0 part

### DIFF
--- a/tests/testapp/test_models.py
+++ b/tests/testapp/test_models.py
@@ -242,8 +242,10 @@ class QueryHintTests(TestCase):
         title_idx = index_name(Book, "title")
         with CaptureLastQuery() as cap:
             list(Author.objects.annotate(has_books=Exists(
-                Book.objects.filter(author_id=OuterRef('id'), title__gt="")
-            )).filter(has_books=True).use_index(title_idx, table_name="testapp_book"))
+                Book.objects.filter(
+                    author_id=OuterRef('id'), title__gt=""
+                ).use_index(title_idx)
+            )).filter(has_books=True))
         assert ("USE INDEX (`" + title_idx + "`)") in cap.query
         used = used_indexes(cap.query)
         assert len(used) == 0 or title_idx in used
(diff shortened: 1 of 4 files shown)


adamchainz:
Thanks! I added a history note and fixed the flake8 error, wil merge and release after CI passes.

hjaarnio:
Hi! Please let me know if a separate ticket for the bug is needed for discussion about the problem / reproduction steps etc

hjaarnio:
Thanks!

adamchainz:
Nothing else is needed, this is clear and well tested, thanks. But if you could also reproduce it in the test suite at the ORM level, that would make the feature more robust.

hjaarnio:
Hi! Added a model level test for this as well. If CI passes everything clear from my side

tests/testapp/test_models.py
```diff
@@ -236,6 +237,22 @@ def test_use_index_primary(self):
...
+                    has_books=Exists(
+                        Book.objects.filter(author_id=OuterRef("id"), title__gt="")
+                    )
+                )
+                .filter(has_books=True)
+                .use_index(title_idx, table_name="testapp_book")
+            )
+        assert ("USE INDEX (`" + title_idx + "`)") in cap.query
+        used = used_indexes(cap.query)
+        assert len(used) == 0 or title_idx in used
```
adamchainz, line 254: Why is there an `OR`? I'd prefer if tests were deterministic

tests/testapp/test_models.py
```diff
@@ -236,6 +237,22 @@ def test_use_index_primary(self):
...
+        title_idx = index_name(Book, "title")
+        with CaptureLastQuery() as cap:
+            list(
+                Author.objects.annotate(
+                    has_books=Exists(
+                        Book.objects.filter(author_id=OuterRef("id"), title__gt="")
+                    )
+                )
+                .filter(has_books=True)
+                .use_index(title_idx, table_name="testapp_book")
```
adamchainz, line 250: why isn't this on the `Book` subquery?

tests/testapp/test_models.py
hjaarnio: It's the same on the two tests above. the assert on line 252 is deterministic, but whether mySQL actually decides to use it in the query plan despite the USE index hint, which is the result of used_indexes, it just seems to run .explain() is not something that can be controlled from the ORM side.

tests/testapp/test_models.py
hjaarnio: for some reason I thought it wouldn't work without table_name in the subquery since the query rewriting works in context of the finished query, but seems like that wasn't correct. will move it

hjaarnio:
Added also focer_index test (that should be deterministic on the query planner level), and moved the hint to the subquery

tests/testapp/test_models.py
adamchainz: Ah yes - I haven't worked on this in years :)

adamchainz:
Released in version 3.12.0, thanks!

PR #883, opened 2022-01-12
Add FixedCharField

le717:
Apologies if something is wrong. I had a time trying to get this project set up on my laptop. 😅

This PR adds a new model `FixedCharField`. It inherits from `django.db.models.CharField` but creates a `char` field, useful for storing fixed string length data. The length is specified through a `length` parameter and disallows passing a `max_length` as it doesn't makes sense in this context. It also checks that the length is within the 0-255 range, as required by MySQL and MariaDB. There is a minor engine implementation difference in that MariaDB defaults to a length of `1` (2) while MySQL doesn't have a default value (3). I have followed the MariaDB behavior.

I'm not at all sure the introduction paragraph is correct at all and very likely needs major correction. 😬

This was requested before in Django itself 13 years ago but was rejected as wontfix. (1).

In case you're wondering about the motivation for this PR, I'm trying to design a model for a legacy table to improve maintenance and there are instances of `char` fields that I would like to accurately map. This project is already helping out with the inclusion of the `EnumField` and I figured why not have this field to the library too. I'm sure I'm not the only one who can benefit from this.

### References
1. https://code.djangoproject.com/ticket/9349
2. https://dev.mysql.com/doc/refman/8.0/en/char.html
3. https://mariadb.com/kb/en/char/

### DIFF
--- a/docs/exposition.rst
+++ b/docs/exposition.rst
@@ -150,7 +150,7 @@ restricted to a set of choices to be stored in a space efficient manner:
 
 
 FixedCharField
----------
+--------------
 
 A field class for using MySQL's ``CHAR`` type, which allows strings to be
 stored at a fixed width:
--- a/docs/exposition.rst
+++ b/docs/exposition.rst
@@ -149,6 +149,20 @@ restricted to a set of choices to be stored in a space efficient manner:
 :ref:`Read more <enum-field>`
 
 
+FixedCharField
+---------
+
+A field class for using MySQL's ``CHAR`` type, which allows strings to be
+stored at a fixed width:
+
+.. code-block:: python
+
+    class Address(Model):
+        zip_code = FixedCharField(length=10)
+
+:ref:`Read more <fixedchar-field>`
+
+
 Resizable Text/Binary Fields
 ----------------------------
 
--- a/docs/model_fields/fixedchar_field.rst
+++ b/docs/model_fields/fixedchar_field.rst
@@ -6,14 +6,12 @@ FixedCharField
 
 .. currentmodule:: django_mysql.models
 
-Using a ``CharField`` for fixed-width strings leads to variable data
-storage since the storage amount can changed depending on the length
-of the string, whereas a fixed-width string will always use the same
-amount of storage, regardless of the string value. Fixed value columns
-can also decrease the storage size overall due to how the engine
-indexes strings. MySQL's ``CHAR`` type allows a more compact representation
-of such columns. ``FixedCharField`` allows you to use the ``CHAR`` type
-with Django.
+Django’s ``CharField`` uses the ``VARCHAR`` data type, which uses variable
+storage space depending on string length. This normally saves storage space,
+but for columns with a fixed length, it adds a small overhead.
+``FixedCharField`` uses the ``CHAR`` data type, which avoids that overhead.
+It only accepts strings with the given length, and uses a corresponding fixed
+amount of storage.
 
 Docs:
 `MySQL <https://dev.mysql.com/doc/refman/en/char.html>`_ /
@@ -34,5 +32,5 @@ Docs:
         from django_mysql.models import FixedCharField
 
 
-        class VariousCharLengths(Model):
+        class Address(Model):
             zip_code = FixedCharField(length=5)
--- a/docs/model_fields/fixedchar_field.rst
+++ b/docs/model_fields/fixedchar_field.rst
@@ -39,10 +39,10 @@ Docs:
         class VariousCharLengths(Model):
             zip_code = FixedCharField(length=5)
             default_length = FixedCharField()  # defaults to length=1
-            really_long_string = FixedCharField(length=256)  # ValueError
+            really_long_string = FixedCharField(length=256)  # raise ValueError
 
-    .. warning::
+    .. note::
 
-    MariaDB defaults to a ``CHAR(1)`` field, while MySQL has no default value.
-    ``FixedCharField`` follows the MariaDB behavior and defaults to a
-    ``CHAR(1)`` field if a length is not provided.
+        MariaDB defaults to a ``CHAR(1)`` field, while MySQL has no default value.
+        ``FixedCharField`` follows the MariaDB behavior and defaults to a
+        ``CHAR(1)`` field if a length is not provided.
--- a/docs/model_fields/fixedchar_field.rst
+++ b/docs/model_fields/fixedchar_field.rst
@@ -1,8 +1,8 @@
 .. fixedchar-field:
 
----------
+--------------
 FixedCharField
----------
+--------------
 
 .. currentmodule:: django_mysql.models
 
@@ -15,11 +15,9 @@ indexes strings. MySQL's ``CHAR`` type allows a more compact representation
 of such columns. ``FixedCharField`` allows you to use the ``CHAR`` type
 with Django.
 
-
 Docs:
 `MySQL <https://dev.mysql.com/doc/refman/en/char.html>`_ /
 `MariaDB <https://mariadb.com/kb/en/char/>`_.
-
 
 .. class:: FixedCharField(length: int, **kwargs)
 
@@ -38,4 +36,3 @@ Docs:
 
         class VariousCharLengths(Model):
             zip_code = FixedCharField(length=5)
-            really_long_string = FixedCharField(length=256)  # raise ValueError
--- a/docs/model_fields/fixedchar_field.rst
+++ b/docs/model_fields/fixedchar_field.rst
@@ -21,7 +21,7 @@ Docs:
 `MariaDB <https://mariadb.com/kb/en/char/>`_.
 
 
-.. class:: FixedCharField(length: int = 1, **kwargs)
+.. class:: FixedCharField(length: int, **kwargs)
 
     A subclass of Django's :class:`~django.db.models.Charfield` that uses a
     MySQL ``CHAR`` for storage.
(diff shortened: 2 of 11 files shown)


adamchainz:
Thanks for the PR. This is the perfect kind of feature for Django-MySQL, and I appreciate that you've put a lot of effort into the PR and documenting in the same way as the rest of the fields.

Could you add a changelog note in `HISTORY.rst`?

I'll give a full review later.

> There is a minor engine implementation difference in that MariaDB defaults to a length of 1 (2) while MySQL doesn't have a default value (3)

Can we just drop the default? IMO confusing to allow it when `CharField` doesn't have a default length. The default is just in the Python function signature too - the field still emits `CHAR(1)` instead of `CHAR` when no length is provided.

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -26,7 +26,10 @@ def __init__(self, length: int, *args: Any, **kwargs: Any) -> None:
         if "max_length" in kwargs:
             raise TypeError('"max_length" is not a valid argument')
 
+        # However, a max_length is required by Django CharField,
+        # so we'll set it equal to the length of our CHAR field
         self.length = length
+        kwargs["max_length"] = length
```
le717, line 32: I considered dropping `self.length` and overloading `max_length` to store the field length and use it to generate the field in `db_type()` while keeping the `length` param but I wasn't sure if that is something you'd want.

docs/exposition.rst
```diff
@@ -149,6 +149,20 @@ restricted to a set of choices to be stored in a space efficient manner:
 :ref:`Read more <enum-field>`
 
 
+FixedCharField
+---------
```
adamchainz, line 153: reStructuredText requires underlines that match the text length above

```suggestion
FixedCharField
--------------
```

docs/model_fields/fixedchar_field.rst
```diff
@@ -0,0 +1,41 @@
+.. fixedchar-field:
+
+---------
+FixedCharField
+---------
```
adamchainz, line 5:
```suggestion
--------------
FixedCharField
--------------
```

docs/model_fields/fixedchar_field.rst
```diff
@@ -0,0 +1,41 @@
...
+Using a ``CharField`` for fixed-width strings leads to variable data
+storage since the storage amount can changed depending on the length
+of the string, whereas a fixed-width string will always use the same
+amount of storage, regardless of the string value. Fixed value columns
+can also decrease the storage size overall due to how the engine
+indexes strings. MySQL's ``CHAR`` type allows a more compact representation
+of such columns. ``FixedCharField`` allows you to use the ``CHAR`` type
+with Django.
+
+
```
adamchainz, line 18: Unnecessary blank line
```suggestion
```

docs/model_fields/fixedchar_field.rst
```diff
@@ -0,0 +1,41 @@
...
+Docs:
+`MySQL <https://dev.mysql.com/doc/refman/en/char.html>`_ /
+`MariaDB <https://mariadb.com/kb/en/char/>`_.
+
+
```
adamchainz, line 23:
```suggestion
```

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,51 @@
+from typing import Any, cast
+
+from django.db.backends.base.base import BaseDatabaseWrapper
+from django.db.models import CharField
+
+from django_mysql.typing import DeconstructResult
+
+
+class FixedCharField(CharField):
+    def __init__(self, length: int, *args: Any, **kwargs: Any) -> None:
```
adamchainz, line 10: Make the arg kwarg only
```suggestion
    def __init__(self, *args: Any, length: int, **kwargs: Any) -> None:
```

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,51 @@
...
+        # Ensure we have an actual integer value
```
adamchainz, line 11: Unnecessary comment, the code is clear
```suggestion
```

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,51 @@
...
+        if not isinstance(length, int):
+            raise TypeError(
+                'Invalid length value "{length}". '
+                "Expected integer value.".format(length=length)
+            )
+
+        # CHAR fields can only be in the range of 0-255
```
adamchainz, line 18: Also clear
```suggestion
```

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,51 @@
...
+        if length < 0 or length > 255:
+            raise ValueError(
+                'Invalid length value "{length}". '
+                "Length must be in the range of 0-255.".format(length=length)
+            )
+
+        # A max_length doesn't make sense in this context
```
adamchainz, line 25:
```suggestion
```

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,51 @@
...
+        if "max_length" in kwargs:
+            raise TypeError('"max_length" is not a valid argument')
+
+        # However, a max_length is required by Django CharField,
+        # so we'll set it equal to the length of our CHAR field
```
adamchainz, line 30:
```suggestion
        # max_length is required by CharField
```

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,51 @@
...
+
+        bad_paths = (
+            "django_mysql.models.fields.fixedchar.FixedCharField",
+            "django_mysql.models.fields.FixedCharField",
+        )
+        if path in bad_paths:
+            path = "django_mysql.models.FixedCharField"
+
+        kwargs["length"] = self.length
+        kwargs["max_length"] = self.length
```
adamchainz, line 46: This will fail the `if "max_length" in kwargs:` on reconstruction?

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,51 @@
...
+
+        return name, path, args, kwargs
+
+    def db_type(self, connection: BaseDatabaseWrapper) -> str:
+        return f"char({self.length})"
```
adamchainz, line 51: Django uses upper case for SQL and this package follows that convention
```suggestion
        return f"CHAR({self.length})"
```

docs/model_fields/fixedchar_field.rst
```diff
@@ -0,0 +1,41 @@
...
+    For example:
+
+    .. code-block:: python
+
+        from django_mysql.models import FixedCharField
+
+
+        class VariousCharLengths(Model):
+            zip_code = FixedCharField(length=5)
+            really_long_string = FixedCharField(length=256)  # raise ValueError
```
adamchainz, line 41: No point showing an invalid example
```suggestion
```

tests/testapp/test_fixedcharfield.py
```diff
@@ -0,0 +1,66 @@
...
+    def test_invalid_length_too_long(self):
+        with pytest.raises(ValueError) as exc_info:
+            FixedCharField(length=256)
+        assert "Length must be in the range" in str(exc_info.value)
+
+    def test_invalid_max_length(self):
+        with pytest.raises(TypeError) as exc_info:
+            FixedCharField(length=4, max_length=100)
+        assert '"max_length" is not a valid argument' in str(exc_info.value)
+
```
adamchainz, line 29: Need tests that actually store data in the field and query it! otherwise there's no proof it works

src/django_mysql/models/fields/fixedchar.py
adamchainz: Indeed the test fails, guess you didn't run it locally

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,49 @@
...
+from django_mysql.typing import DeconstructResult
+
+
+class FixedCharField(CharField):
+    def __init__(self, *args: Any, length: int, **kwargs: Any) -> None:
+        if not isinstance(length, int):
+            raise TypeError(
+                'Invalid length value "{length}". '
+                "Expected integer value.".format(length=length)
+            )
```
adamchainz, line 17: This check is unnecessary - we already have both the type hint and the system check in `CharField`:

https://github.com/django/django/blob/7c4f3965098baad2396e24501e09237425a7bd6f/django/db/models/fields/__init__.py#L1020

src/django_mysql/models/fields/fixedchar.py
adamchainz: Yes this seems like a good idea

src/django_mysql/models/fields/fixedchar.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if length < 0 or length > 255:
+            raise ValueError(
+                'Invalid length value "{length}". '
+                "Length must be in the range of 0-255.".format(length=length)
+            )
```
adamchainz, line 23: I think we should follow Django's example and perform this in a system check. And use an f string.

adamchainz:
Okay, whilst writing tests I realized what I hadn't read before - `CHAR` adds trailing spaces on storage, and (normally) strips them on retrieval. So the 'length' really acts more like a 'max length'. Therefore I'll be renaming the argument to just `max_length`.

adamchainz:
Thank you for the contribution. I've just released it in version 4.5.0 .

le717:
Thank you for considering this! I apologize for all the errors. I had a lot of trouble trying to get anything running on my Windows machine (from installation to running tests) so I basically had to do everything blind and using the other model fields as reference to set up tests and everything else.

adamchainz:
No worries, thanks for the contribution.

PR #948, opened 2022-09-14
Add support for filter expression in GroupConcat

caramdache:
GroupConcat did not support fitler expression. This PR adds support based on Django `Count` aggregate.

https://github.com/adamchainz/django-mysql/blob/main/src/django_mysql/models/aggregates.py

### DIFF
--- a/src/django_mysql/models/aggregates.py
+++ b/src/django_mysql/models/aggregates.py
@@ -28,6 +28,7 @@ class GroupConcat(Aggregate):
     def __init__(
         self,
         expression: Expression,
+        filter = None,
         distinct: bool = False,
         separator: str | None = None,
         ordering: str | None = None,
@@ -38,7 +39,7 @@ class GroupConcat(Aggregate):
             # This can/will be improved to SetTextField or ListTextField
             extra["output_field"] = CharField()
 
-        super().__init__(expression, **extra)
+        super().__init__(expression, filter=filter, **extra)
 
         self.distinct = distinct
         self.separator = separator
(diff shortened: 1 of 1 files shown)


src/django_mysql/models/aggregates.py
```diff
@@ -28,6 +28,7 @@ class GroupConcat(Aggregate):
     def __init__(
         self,
         expression: Expression,
+        filter=None,
```
adamchainz, line 31: We require type hints. We can copy from django-stubs here ( https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/models/aggregates.pyi ) and just use `Any | None`, which isn't much of a type hint, but if django-stubs gets better later, we can copy it.

```suggestion
        filter: Any | None=None,
```

src/django_mysql/models/aggregates.py
```diff
@@ -53,6 +54,15 @@ def as_sql(
...
         **extra_context: Any,
     ) -> tuple[str, tuple[Any, ...]]:
+        if self.filter:
+            extra_context["distinct"] = "DISTINCT " if self.distinct else ""
+            copy = self.copy()
+            copy.filter = None
+            source_expressions = copy.get_source_expressions()
+            condition = When(self.filter, then=source_expressions[0])
+            copy.set_source_expressions([Case(condition)] + source_expressions[1:])
+            return super(Aggregate, copy).as_sql(compiler, connection, **extra_context)
```
adamchainz, line 64: I haven't looked into it, but it's a bit odd that you're skipping the existing implementation and calling `super()`. The implemetnation below covers the SQL oddities in `GROUP_CONCAT`, I think we'd still want to use it, just tweak adding the filter there.

adamchainz (changes requested):
Nice work starting on this.

The tests are failing because the package requires test coverage - please add tests in the existing tests for group concat. It looks like we'll need at least two, depending on the value of `distinct`.

Also please update the documentation the changelog.

src/django_mysql/models/aggregates.py
caramdache: I'm not sure we need keep the existing pathway in the long term. It's probably better to let `Aggregate.as_sql()` perform the work based on the template.

If you look at the code for `Count`, you'll see there are 2 paths, one when filter is set, the other when it's not. I think we should have a similar structure here, but avoid manual handcrafting as much as possible because this is likely fragile.

adamchainz:
THat's the file, but you've done so in a separate PR: #949. Please do so in *this* PR, on the same branch.

adamchainz:
Still waiting on docs, changelog note, and tests.

jvacek:
@caramdache Any status updates? Wondering if I can help implement this somehow without having to start a whole new MR.

adamchainz:
I think it's unlikely they'll return, please start a new PR. You can pull their changes with `gh pr checkout 948` then create a new branch and [squash rebase](https://adamj.eu/tech/2022/03/25/how-to-squash-and-rebase-a-git-branch/) on top of the latest `main` before adding your own updates.

PR #949, opened 2022-09-14
Add support for ``GroupConcat`` filters.

caramdache:


adamchainz:
presuming this was an accident, please do in #948.

PR #951, opened 2022-09-20
Add model TinyIntegerField/PositiveTinyIntegerField

le717:
This starts adding support for the `TINYINT` field, with an unsigned version available too.

Building proper model fields (i.e., not simply subclassing `IntegerField` and overriding `db_type`) is still very foreign to me, so I appreciate any and all suggestions to help me learn how to better make these. 😃 

### TODO
- [ ] Confirm actual construction is correct
- [x] Add test model/migration
- [ ] Add/run tests
- [ ] Add documentation
- [ ] Any feedback you have

### Documentation
- https://mariadb.com/kb/en/tinyint/
- https://dev.mysql.com/doc/refman/en/integer-types.html

### DIFF
--- a/src/django_mysql/models/__init__.py
+++ b/src/django_mysql/models/__init__.py
@@ -25,4 +25,6 @@ from django_mysql.models.fields import (  # noqa
     SetTextField,
     SizedBinaryField,
     SizedTextField,
+    TinyIntegerField,
+    PositiveTinyIntegerField,
 )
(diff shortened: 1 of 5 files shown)


adamchainz:
I guess it would be good to add this for completeness, but I'd like to understand a bit more about why you'd like such a field type. It's best to add issues before working on features, to provide context. Everything added to the package will need maintaining until the heat death of the universe.

Also workflow-wise, it'd be best if you could set up [pre-commit](https://pre-commit.com/) locally, so the CI service doesn't need to push fix commits.

PR #970, opened 2022-11-30
Fix set fields’ contains lookup

tuky:
This should be obvious enough to merge and ship quickly, I hope. Thx!

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,8 @@ History
 * Make ``MySQLCache.touch()`` return ``True`` if the key was touched, ``False`` otherwise.
   This return value was missing since the method was added for Django 2.1.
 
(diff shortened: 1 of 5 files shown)


tuky:
Funny enough this bug should be 8 years old now and only emerged for us with django 4.1. I assume they refactored / reordered some query building logic and with that query parts got split into lhs and rhs.

adamchainz:
Please add a regression test and a release note.

tuky:
ah, great stuff with the auto fixing :) :+1:

adamchainz:
I wrote a simpler test that checked the lookup directly on the field, and copied it between the two field variants. Also changed release note to document things at the level of the project (the set fields) rather than the underlying SQL funciton name (kind of an implementation detail).

adamchainz:
Released in 4.8.0, thanks!

tuky:
thx a lot!

PR #1018, opened 2023-06-23
Added support for NATURAL_SORT_KEY with MariaDB

### LINKED ISSUE #975: Add support for natural sorting with MariaDB 10.7+ `NATURAL_SORT_KEY`
### Description

It would be great to be able to sort by [NATURAL_SORT_KEY](https://mariadb.com/kb/en/natural_sort_key/).

Possibly the following could be supported for the syntax (to natsort by the `name` column):

```python
Document.objects.order_by("NATURAL_SORT_KEY(name)")
```

Thanks

aduuna:
Fixes #975

This PR adds support for MariaDB's NATURAL_SORT_KEY. It has been implemented as a database function.

```Alphabet.objects.order_by(NaturalSortKey("d"))```

### DIFF
--- a/src/django_mysql/models/functions.py
+++ b/src/django_mysql/models/functions.py
@@ -129,7 +129,7 @@ class NaturalSortKey(Func):
         self,
         compiler: SQLCompiler,
         connection: BaseDatabaseWrapper,
-        **extra_context,
+        **extra_context: dict[str, Any],
     ) -> tuple[str, tuple[Any, ...]]:
         if connection.vendor == "mysql" and not connection.mysql_is_mariadb:
             raise AssertionError("NATURAL_SORT_KEY is not supported by MySQL")
--- a/src/django_mysql/models/functions.py
+++ b/src/django_mysql/models/functions.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime as dt
 import json
+from packaging.version import Version
 from typing import Any
 from typing import Union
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by adamchainz — CHANGES_REQUESTED** [review]
I'm not interested in adding database functions that take so little effort to implement in your own project. The functions we have in the package have more restricted types, which helps users call them correctly. Could you do the same here?

Also every pull request needs documentation and a changelog note.

File: src/django_mysql/models/functions.py [hunk 1]
```diff
@@ -118,6 +119,25 @@ def __init__(
...
+    def __init__(self, expression: ExpressionArgument) -> None:
+        super().__init__(expression)
+
+    def as_sql(
+        self,
+        compiler: SQLCompiler,
+        connection: BaseDatabaseWrapper,
+        **extra_context,
+    ) -> tuple[str, tuple[Any, ...]]:
+        if connection.vendor == "mysql" and not connection.mysql_is_mariadb:
```
adamchainz, line 134: These database version checks are not worth it. In the successful path, all of this checking is time wasted (creating a `Version` is not free!). In the incredibly rare failure path, they’re not really providing more information than the database’s own error (“no such function”) would provide.

This is why none of the other functions have such checks.

adamchainz:
I’m sorry I still don’t think we are adding much of utility above the original two line function definition. Simply restricting to a single expression isn’t reason enough to add a custom class and maintain it forever. Thanks.

src/django_mysql/models/functions.py
aduuna: Thank you for your feedback. I understand your concerns about the database version checks in the code. But my reasoning was that While these checks can introduce overhead and may not always provide significant additional information, they serve as a preventive measure to catch potential issues early on. 

I will look through the open issue pool on the Django projects to find a first-time-contributor friendly to contribute to.

adamchainz: Yeah, I'm all for preventative checks, but their cost has to be balanced against the probability of making a given mistake.

PR #1084, opened 2024-04-25
Lazily import django.test to reduce import time

angusholder:
I was running a flamegraph on our manage.py, and noticed this import of `django.test` using 20-40ms. Seems like an easy win for startup time to make it a lazy import, because most production code wouldn't normally ever need the `django.test` module.

### DIFF
--- a/src/django_mysql/models/query.py
+++ b/src/django_mysql/models/query.py
@@ -23,7 +23,6 @@ from django.db import connections
 from django.db import models
 from django.db.models.sql.where import ExtraWhere
 from django.db.transaction import atomic
-from django.test.utils import CaptureQueriesContext
 from django.utils.functional import cached_property
 from django.utils.translation import gettext as _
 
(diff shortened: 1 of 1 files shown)


adamchainz:
Thanks, that is an easy win. I added a changelog note and improved the comment.

This function could be improved further to not use `django.test` and instead capture the query with [database instrumentation](https://docs.djangoproject.com/en/5.0/topics/db/instrumentation/), if you’re up for the challenge.

angusholder:
After sending this, I realised it was _your_ blog posts I read recently about improving Django start time, so thanks for those!

Might have a go at that database instrumentation change later, I haven't seen it before, looks interesting.

adamchainz:
Glad you enjoyed them!

I also have some blog posts on database instrumentation that may inform: https://duckduckgo.com/?t=ffab&q=site%3Aadamj.eu+%22database+instrumentation%22&ia=web

The suggestion is just an “if you’re interested”, I don't use `pt_visual_explain` myself, at the moment, and I guess you don’t either :)

PR #1123, opened 2024-12-01
Add tiny integer field

### LINKED ISSUE #1121: Add `TinyIntegerField` and `PositiveTinyIntegerField`
### Description

Hi Adam,

This is a small feature request about adding support for the mysql type `TYNYINT` with 2 fields `TinyIntegerField` and `PositiveTinyIntegerField`. I would be happy to provide a PR for this.

I found a related discussion in #749, however I am unsure this belongs as a Django core feature for the `TINYINT` case because postgresql (which I believe is the most used django DB backend) doesn't have a `TINYINT` type, only the 2 bytes `SMALLINT` type.

Thanks in advance

UnknownPlatypus:
I went for a tiny wrapper on top of  `SmallIntegerField`, its convenient because we want the same implementation for a lot of hooks (`db_check`, `cast_db_type`, `db_type_suffix`) but currently, the `validators` hook is a bit incorrect because it uses `SmallIntegerField` ranges (-32768, 32767) instead of (-128, 127). 

The ranges are stored on `connection.ops.integer_field_range` which makes them a bit difficult to change.
My only working idea for now would be to override `validators` with mostly the `IntegerField.validators` hook implementation, just altering the min/max values but it's not very clean and might drift.

Do you have a recommendation on how to solve this ? 

Fixes #1121


adamchainz:
> The ranges are stored on `connection.ops.integer_field_range` which makes them a bit difficult to change. My only working idea for now would be to override `validators` with mostly the `IntegerField.validators` hook implementation, just altering the min/max values but it's not very clean and might drift.
> 
> Do you have a recommendation on how to solve this ?

I looked a bit more at Django’s source and found `Field.default_validators` can be used to fix this. I just pushed a commit doing so.

adamchainz:
Thanks for this! I did a bunch of small cleanup, which you might want to check the commits for.

PR #1137, opened 2025-03-20
force_index and use_index PRIMARY clarify

akvilibrium:
Clarified PRIMARY usage for indexes hints


adamchainz:
I do not understand why you made this change. Please explain it in the PR description and add a changelog note.

Issue #1114, opened 2024-09-20
Add way to support Column Prefix Key Parts


tuky:
### Description

Heyo!

See my ticket https://code.djangoproject.com/ticket/35777. It seems to me, that django lacks support to create indexes with so-called "Column Prefix Key Parts". These are documented here: https://dev.mysql.com/doc/refman/8.4/en/create-index.html#create-index-column-prefixes and give you a huge performance boost, because these indexes can then be used for `istartswith` queries on `TEXT` fields. Would this be something for django itself or should this be rather implemented here? How would you evaluate my workaround from the django ticket?

```python
class ColumnPrefixIndex:
    max_name_length = 60
    contains_expressions = False
    fields_orders = []
    include = []
    expressions = []
    fields = []
    condition = None

    def __init__(self, column, prefix, name):
        self.prefix = prefix
        self.column = column
        self.name = name

    def create_sql(self, model, schema_editor, using="", **kwargs):
        table = model._meta.db_table
        sql = f'CREATE INDEX {self.name} ON {table} ({self.column}({self.prefix}));'
        kwargs['sql'] = sql

        return schema_editor._create_index_sql(
            model,
            name=self.name,
            using=using,
            **kwargs,
        )

    def deconstruct(self):
        kwargs = {'column': self.column, 'prefix': self.prefix, 'name': self.name}
        path = "%s.%s" % (self.__class__.__module__, self.__class__.__name__)
        return path, self.expressions, kwargs

    def clone(self):
        return self.__class__(self.column, self.prefix, self.name)
```

adamchainz:
Hi!

Yeah, we could add such an index class here. It seems to be MySQL/MariaDB specific, so less suitable for Django core, which only has `django.contrib.postgres`.

Rough plan to get it ready for merging:

1. Add to `django_mysql.models.indexes`.
2. Add type hints.
3. Add tests. Some of the existing field and operations tests can be used for inspiration. We have 100% coverage here!
4. Add docs - a new page called "indexes" covering it, with the same level of detail as existing docs.
5. Add mention in "exposition" page of docs.

If you have time for this, that would be great!

PR #1151, opened 2025-06-23
Add ColumnPrefixIndex for MySQL column prefix key parts

JaeHyuckSa:
## Description
This PR implements ColumnPrefixIndex to support MySQL's Column Prefix Key Parts feature. This feature can significantly improve performance for istartswith queries on TEXT or long VARCHAR fields while being space-efficient.

Related issue: #1114

## Example Usage
```python
class Article(models.Model):
    title = models.CharField(max_length=200)
    content = models.TextField()

    class Meta:
        indexes = [
            ColumnPrefixIndex(
                fields=['title', 'content'],
                prefix_lengths=(10, 50),
                name='title_content_prefix_idx'
            ),
        ]
```

### DIFF
--- a/docs/exposition.rst
+++ b/docs/exposition.rst
@@ -199,7 +199,7 @@ field class allows you to interact with those fields:
 Tiny integer fields
 -------------------
 
-MySQL’s ``TINYINT`` type stores small integers efficiently, using just one byte.
+MySQL's ``TINYINT`` type stores small integers efficiently, using just one byte.
 Django-MySQL provides field classes for the ``TINYINT`` type:
 
 .. code-block:: python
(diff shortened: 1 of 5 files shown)


adamchainz:
Hi

On a second reading, I don't think this should be a separate index class, but instead an expression/function.

MySQL allows indexes to mix prefixes as well as full column indexes, like: `CREATE INDEX example_idx ON example (title(10), state)`. To support this, an expression should allow us to do something like:

```py
Index(
   Prefix("title", 10),
   "state",
   ...,
)
```

Can you investigate that path instead?

JaeHyuckSa:
@adamchainz You mean that in MySQL’s `_create_index_sql`, whenever it encounters a `Slice` expression for an index, it should emit the `column(n))` syntax instead of `SUBSTRING(col, 1, n)`, correct?

